PR #105, opened 2021-01-27
Login and Register

nienna73:
You can now register a new user then login as that user. There are also tests for the login/register UI and basic error checking. Logging in doesn't do anything for you at this time. Meaning all the functionality is available whether or not you're logged in.


.gitignore
```diff
@@ -112,3 +112,5 @@ ENV/
 metadata.csv
 /data/
 db.sqlite3
+node_modules
+.DS_Store
```
andrewdotn, line 116: `cypress/.DS_Store` snuck into the PR despite this

validation/views.py
```diff
@@ -153,4 +163,55 @@ def add_cors_headers(response):
...
+                response = HttpResponseRedirect("/")
+                return response
+    form = Login()
+    context = dict(form=form)
+    return render(request, "validation/login.html", context)
+
+
+def register(request):
+    """
+    Serves the login page and sets cookies on successful login
```
andrewdotn: Copy-paste of docstring from previous method?

validation/forms.py
```diff
@@ -0,0 +1,62 @@
...
+        required=True,
+        widget=forms.PasswordInput(
+            attrs={"class": "form-control", "style": "max-width: 15rem;"}
+        ),
+    )
+
+    def clean_username(self):
+        username = self.cleaned_data["username"]
+        if not re.search(
+            r"^\w+$", username
```
andrewdotn, line 51: I had to look this up, but `\w` matches alphanumerics in any Unicode script, e.g., `î४`. That’s probably fine? I just found it interesting.

validation/views.py
```diff
@@ -153,4 +163,55 @@ def add_cors_headers(response):
     return response
 
 
+def login(request):
+    """
+    Serves the login page and sets cookies on successful login
+    """
+
+    if request.method == "GET":
```
andrewdotn: Login should use the `POST` method. Among other things, this would record user’s passwords in the server log. Other than that it could be ok, but you may want to use the built-in [LoginView](https://docs.djangoproject.com/en/3.1/topics/auth/default/#django.contrib.auth.views.LoginView)

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "cypress": "^6.3.0"
```
andrewdotn, line 3: Yay! Having cypress tests is great!

validation/views.py
nienna73: Good catch!

.gitignore
nienna73: Oops. It _should_ be gone now.

validation/forms.py
nienna73: I'm not sure that we have any requirements for username and I highly doubt our users will be using special characters in their usernames. I do wonder if this would cause issues in how the usernames are stored.

validation/views.py
nienna73: Excellent suggestion! I ideally don't want passwords being stored in plain text, so I'll look into the LoginView.

cypress.json
```diff
@@ -0,0 +1,6 @@
+{
+    "baseUrl": "http://localhost:8000",
+    "env": {
+        "login_url": "/login"
```
eddieantonio: Should you be using this in the tests as `Cypress.env()`?

validation/jinja2/validation/_base.html
```diff
@@ -13,6 +13,10 @@
     margin: 1em auto;
   }
 
+  .bg-dark {
+    background-color: #143B6D !important;
```
eddieantonio, line 17: What kind of specificity conflict requires the use of `!important`?

validation/jinja2/validation/_base.html
```diff
@@ -54,6 +98,16 @@
         <button class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
       </form>
     </div>
+
+    {# Login button #}
+    <div>
+      <a href="/login" style="color: white;">
```
eddieantonio: Extract inline style as a class?

eddieantonio (commented):
Just a few questions and suggestions, but I'm excited!

cypress.json
nienna73: Yes, yes I should. The way I'm calling it right now just appends the string to the `baseUrl`. I'll change it to use `Cypress.env()`

validation/jinja2/validation/_base.html
nienna73: `.bg-dark` is a bootstrap styles class that I wanted to change a little. There will be another PR at the end of all this with properly implemented styles in their own file.

validation/jinja2/validation/_base.html
nienna73: Most definitely. This will also be handled in the styles PR to come.

.gitignore
andrewdotn: Oh, the original `.DS_Store` line in `.gitignore` was just fine. I think what happened was that when the `.gitignore` line was added, that was after `.DS_Store` was already `git add`ed, even if not yet `git commit`ed.

validation/views.py
```diff
@@ -153,4 +163,52 @@ def add_cors_headers(response):
...
+                    last_name=last_name,
+                )
+                new_user.save()
+                response = HttpResponseRedirect("/login")
+                return response
+
+    context = dict(form=form)
+    return render(request, "validation/register.html", context)
+
+
```
andrewdotn, line 213: These are unused now, right?

```suggestion
```

validation/views.py
```diff
@@ -153,4 +163,52 @@ def add_cors_headers(response):
...
+            user = authenticate(request, username=username, password=password)
+            if user is not None:
+                django_login(request, user)
+                response = HttpResponseRedirect("/")
+                return response
+    form = Login()
+    context = dict(form=form)
+    return render(request, "validation/login.html", context)
+
+
```
andrewdotn: This method is unused now, right?
```suggestion
```

validation/views.py
```diff
@@ -153,4 +163,52 @@ def add_cors_headers(response):
...
+def register(request):
+    """
+    Serves the register page and creates a new user on success
+    """
+
+    if request.method == "GET":
+        form = Register(request.GET)
```
```suggestion
    if request.method == "POST":
        form = Register(request.POST)
```

validation/jinja2/validation/register.html
```diff
@@ -0,0 +1,16 @@
+{% extends 'validation/_base.html' %}
+{% import 'validation/_macros.html' as macros %}
+
+
+{% block content %}
+<div class="login">
+<h2>Register as a New User</h2>
+    <div class="login-form-wrapper">
+        <form action="" id="register-form" method="GET">
```
andrewdotn:
```suggestion
        <form action="" id="register-form" method="POST">
```

.gitignore
```diff
@@ -112,3 +112,6 @@ ENV/
 metadata.csv
 /data/
 db.sqlite3
+node_modules
+*.DS_Store
+*/.DS_Store
```
eddieantonio:
```suggestion
.DS_Store
```

(see: https://github.com/github/gitignore/blob/master/Global/macOS.gitignore)

cypress.json
```diff
@@ -0,0 +1,8 @@
+{
+    "baseUrl": "http://localhost:8000",
+    "env": {
+        "login_url": "/login",
+        "register_url": "/register",
+        "logout_url": "/logout"
+      }
```
eddieantonio, line 7: 🤩

validation/views.py
nienna73: OH. You're correct, consider it gone.

PR #106, opened 2021-01-29
Page for phrase

nienna73:
This PR has the details view for each phrase. I accidentally branched off my initial changes from trying to extract ELAN data, so some files have been reformatted as a result of me removing those changes.

This is the first PR that uses divvunspell, so I would appreciate it if you could test it locally as well as reading through the code for errors.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -15,6 +15,7 @@ python-decouple = "*"
 sh = "*"
 django = ">=2.2,<3"
 typing-extensions = "*"
+pympi-ling = "*"
 
 [dev-packages]
 black = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -16,6 +16,7 @@ sh = "*"
 django = ">=2.2,<3"
 typing-extensions = "*"
 pympi-ling = "*"
+requests = "*"
 
 [dev-packages]
 gunicorn = "*"
@@ -30,4 +31,3 @@ django-extensions = "*"
 
 [requires]
 python_full_version = "3.7.9"
-
--- a/Pipfile
+++ b/Pipfile
@@ -18,7 +18,6 @@ typing-extensions = "*"
 pympi-ling = "*"
 
 [dev-packages]
-black = "*"
 gunicorn = "*"
 hypothesis = "*"
 isort = ">=5,<6"
@@ -30,7 +29,5 @@ django-stubs = "*"
 django-extensions = "*"
 
 [requires]
-python_version = "3.7"
+python_full_version = "3.7.9"
 
-[pipenv]
-allow_prereleases = true
--- a/Pipfile
+++ b/Pipfile
@@ -17,6 +17,7 @@ django = ">=2.2,<3"
 typing-extensions = "*"
 pympi-ling = "*"
 requests = "*"
+divvunspell-python = {ref = "3aa656d36a2fa1ede90e53c320b2095c05cf7c05", git = "https://github.com/divvun/divvunspell-sdk-python"}
 
 [dev-packages]
 gunicorn = "*"
--- a/validation/helpers.py
+++ b/validation/helpers.py
@@ -193,8 +193,3 @@ def get_distance_with_translations(word):
         }
 
     return suggestions
-
-
-if __name__ == "__main__":
-    word = "ekwa"
-    print(get_distance_with_translations(word))
--- a/validation/helpers.py
+++ b/validation/helpers.py
@@ -1,0 +1,195 @@
+import divvunspell
+from difflib import Differ, SequenceMatcher, get_close_matches
+import operator
+
+import requests
+import json
+
+"""
+RULES
+
+    adding/removing diacritics or hyphens, swapping glides w/y, or 
+    adding/removing aspirations -h- between vowel and consonant would have a weight of zero.
+    inserting the vowel -i- between two consonants would have a half-weight (0.5).
+    inserting/removing/swapping any other characters would have a normal weight of one.
+
+"""
+
+archive = divvunspell.SpellerArchive("/Users/jolenepoulin/Documents/crk.zhfst")
+speller = archive.speller()
+
+vowels = ["a", "i", "o", "â", "î", "ô"]
+consonants = ["  c", "  h", "  k", "  m", "  n", "  p", "  s", "  t", "  w", "  y"]
+
+
+def get_edit_distance(word):
+
+    ranked_suggestions = speller.suggest(word)
+    suggestions = [s[0] for s in ranked_suggestions]
+
+    rankings = {}
+    for s in suggestions:
+        med = get_differ(word, s)
+        rankings[s] = med
+
+    sorted_tuples = sorted(rankings.items(), key=operator.itemgetter(1))
+    sorted_rankings = {k: v for k, v in sorted_tuples}
+
+    return sorted_rankings
+
+
+def get_differ(word, suggestion):
+    med = 0
+    dif = Differ()
+    df = list(dif.compare(word, suggestion))
+    for i in range(len(df)):
+        c = df[i]
+
+        # check for replaced e with ê
+        # we should never replace ê with e though
+        if c == "- e":
+            if i + 1 < len(df):
+                if df[i + 1] == "+ ê":
+                    continue
+
+        if c == "+ ê":
+            if i - 1 >= 0:
+                if df[i - 1] == "- e":
+                    continue
+
+        # no change, carry on
+        if "+" not in c and "-" not in c:
+            continue
+
+        # added or removed a hyphen
+        if c == "- -" or c == "+ -":
+            continue
+
+        # check for swapping glides
+        if c == "- y":
+            if i + 1 < len(df):
+                if df[i + 1] == "+ w":
+                    continue
+
+        if c == "- w":
+            if i + 1 < len(df):
+                if df[i + 1] == "+ y":
+                    continue
+
+        if c == "+ y":
+            if i - 1 >= 0:
+                if df[i - 1] == "- w":
+                    continue
+
+        if c == "+ w":
+            if i - 1 >= 0:
+                if df[i - 1] == "- y":
+                    continue
+
+        # end check for swapping glides
+
+        if c.split()[1] in vowels:
+            # adding or removing a diacritic from a vowel has cost 0
+            alt = get_analog(c)
+            if i + 1 < len(df):
+                if df[i + 1] == alt:
+                    continue
+            if i - 1 >= 0:
+                if df[i - 1] == alt:
+                    continue
+
+        if "i" in c:
+            # adding an -i- between two consonants has cost 0.5
+            j = i + 1 if i + 1 < len(df) else None
+            k = i - 1 if i - 1 >= 0 else None
+            if j and k:
+                if df[j] in consonants and df[k] in consonants:
+                    med += 0.5
+                    continue
+
+        if "h" in c:
+            # did we just add an -h- between a vowel and a consonant?
+            j = i + 1 if i + 1 < len(df) else None
+            k = i - 1 if i - 1 >= 0 else None
+            if j and k:
+                after = df[j].replace("-", "")
+                after = after.replace("+", "")
+                before = df[k].replace("-", "")
+                before = before.replace("+", "")
+                after = "  " + after.strip()
+                before = before.strip()
+
+                if after in consonants and before in vowels:
+                    # we did, this costs nothing
+                    continue
+
+        # everything else has cost 1
+        med += 1
+
+    return med
+
+
+def get_analog(char):
+    # take in '- vowel'
+    # return '+ vowel with hat'
+    # OR
+    # take in '+ vowel'
+    # return '- vowel with hat'
+
+    alt = ""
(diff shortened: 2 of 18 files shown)


validation/helpers.py
```diff
@@ -0,0 +1,195 @@
...
+    inserting/removing/swapping any other characters would have a normal weight of one.
+
+"""
+
+
+archive = divvunspell.SpellerArchive(fspath(settings.BASE_DIR / "crk.zhfst"))
+speller = archive.speller()
+
+vowels = ["a", "i", "o", "â", "î", "ô"]
+consonants = ["  c", "  h", "  k", "  m", "  n", "  p", "  s", "  t", "  w", "  y"]
```
andrewdotn: Minor suggestion: I found myself double-checking that there were two spaces in front of every consonant. This could be made explicit with:

    consonants = [f"  {char}" for char in "chkmnpstwy"]

validation/helpers.py
```diff
@@ -0,0 +1,195 @@
...
+    sorted_tuples = sorted(rankings.items(), key=operator.itemgetter(1))
+    sorted_rankings = {k: v for k, v in sorted_tuples}
+
+    return sorted_rankings
+
+
+def get_differ(word, suggestion):
+    med = 0
+    dif = Differ()
+    df = list(dif.compare(word, suggestion))
```
andrewdotn: Since `dif` is never used elsewhere, the code might be cleaner as

```suggestion
    df = list(Differ().compare(word, suggestion))
```

instead of having two variables right next to each other with very similar names

validation/helpers.py
```diff
@@ -0,0 +1,195 @@
...
+    elif char.split()[1] in ["â", "î", "ô"]:
+        alt = alt.replace("â", "a")
+        alt = alt.replace("î", "i")
+        alt = alt.replace("ô", "o")
+
+    return alt
+
+
+def get_translations_from_itwewina(word):
+    url = "https://sapir.artsrn.ualberta.ca/cree-dictionary/click-in-text/?q=" + word
```
andrewdotn: - This URL should probably go into a django settings variable
  - It probably doesn’t matter too much here, but in general it’s a lot safer to use a library function to create query strings. For an example of where things might break otherwise, consider:

        >>> urlencode({'q': 'foo/bar&baz'})
        'q=foo%2Fbar%26baz'

```suggestion
    url = urljoin(settings.ITWEWINA_URL, 'click-in-text/') + '?' + urlencode({'q': word})
```

Though that’ll need this at the top:

    from urllib.parse import urljoin

    from django.conf import settings
    from django.utils.http import urlencode

validation/helpers.py
```diff
@@ -0,0 +1,195 @@
...
+    df = list(dif.compare(word, suggestion))
+    for i in range(len(df)):
+        c = df[i]
+
+        # check for replaced e with ê
+        # we should never replace ê with e though
+        if c == "- e":
+            if i + 1 < len(df):
+                if df[i + 1] == "+ ê":
+                    continue
```
andrewdotn: I know this is getting into nitpick territory, so feel free not to do this, but I have a strong urge to change every `if i ± ? < len: \n if df[i ± ?] == "?"` in this function into a single-line `and` condition, or to add a nested function before the start of the for loop

    def diff_char_or_none(index):
        if index >= 0 and index < len(df):
            return df[index]
        return None

and turn every conditional into something like

    if c == "- e" and diff_char_or_none(i + 1) == "+ ê":
        continue

Pipfile
```diff
@@ -15,9 +15,11 @@ python-decouple = "*"
 sh = "*"
 django = ">=2.2,<3"
 typing-extensions = "*"
+pympi-ling = "*"
+requests = "*"
+divvunspell-python = {ref = "3aa656d36a2fa1ede90e53c320b2095c05cf7c05", git = "https://github.com/divvun/divvunspell-sdk-python"}
 
 [dev-packages]
-black = "*"
```
andrewdotn, line 20: I too get the dreaded `Locking failed … pkg_resources.VersionConflict` message trying to add `black` on this branch, so 👍 on dropping it for now.

Based on [this thread](https://github.com/pypa/pipenv/issues/4476),
`pip install -U virtualenv` of all things *seemed* to have let it install for me, but leaves me even more confused than before.

validation/urls.py
```diff
@@ -51,4 +51,5 @@
         views.all_recordings_for_session,
         name="crude_recordings",
     ),
+    path("<str:segment_id>", views.segment_content_view, name="segment_detail"),
```
andrewdotn:
```suggestion
    path("segment/<str:segment_id>", views.segment_content_view, name="segment_detail"),
```
Otherwise the log fills up with error messages about `/favicon.ico` not parsing as an integer

validation/jinja2/validation/_macros.html
```diff
@@ -59,13 +59,25 @@
 
 {#
  # Expands to a <td> that displays the Cree transcription.
+  # Creates link from phrase to view more details 
  #}
-{% macro td_transcription(user, phrase) %}
+{% macro td_transcription_link(user, phrase) %}
 <td lang="crk">
-  {{ phrase.transcription }}
+  <a name="word-link" id="{{ phrase.transcription }}" href="/{{ phrase.id }} ">
```
andrewdotn:
```suggestion
  <a name="word-link" id="{{ phrase.transcription }}" href="{{ url('validation:segment_detail', phrase.id) }} ">
```

validation/helpers.py
```diff
@@ -0,0 +1,195 @@
...
+        alt = alt.replace("â", "a")
+        alt = alt.replace("î", "i")
+        alt = alt.replace("ô", "o")
+
+    return alt
+
+
+def get_translations_from_itwewina(word):
+    url = "https://sapir.artsrn.ualberta.ca/cree-dictionary/click-in-text/?q=" + word
+    r = requests.get(url, verify=False)
```
andrewdotn:
```suggestion
    r = requests.get(url)
```

Sapir was misconfigured and I updated it. If you still have issues we might need the `certifi` package in the Pipenv. The quick hack fix was to paste the correct intermediate certificate onto sapir. cc: @eddieantonio 

If anyone is interested, I can explain this all in great depth, but am going to assume other people are happy they don’t have to think about it.

```
--- /etc/apache2/sites-available/cree-dictionary-python-3.9.orig	2021-01-29 10:24:01.059058416 -0700
+++ /etc/apache2/sites-available/cree-dictionary-python-3.9.conf	2021-01-29 10:22:02.592036453 -0700
@@ -57,7 +57,7 @@
 #   the referenced file can be the same as SSLCertificateFile
 #   when the CA certificates are directly appended to the server
 #   certificate for convinience.
-SSLCertificateChainFile /etc/ssl/certs/intermediate.pem
+SSLCertificateChainFile /etc/ssl/certs/GlobalSign_RSA_OV_SSL_CA_2018.pem
 
 #   Certificate Authority (CA):
 #   Set the CA certificate verification path where to find CA
```

validation/views.py
```diff
@@ -163,6 +164,19 @@ def add_cors_headers(response):
...
 
+def segment_content_view(request, segment_id):
+    """
+    The view for a single segment
+    Returns the selected phrase and info provided by the helper functions
+    """
+    phrases = Phrase.objects.filter(id=segment_id)
+    segment_name = phrases[0].transcription
+    suggestions = get_distance_with_translations(segment_name)
+    context = dict(phrases=phrases, segment_name=segment_name, suggestions=suggestions)
```
andrewdotn: I think there is a standard way for pages to get auth information but I don’t remember it off-hand. That would be great; this will work for now, and won’t show the “Login” link when already logged in.

```suggestion
    auth = request.user.is_authenticated
    context = dict(phrases=phrases, segment_name=segment_name, suggestions=suggestions, auth=auth)
```

validation/helpers.py
nienna73: Yes, this is so much cleaner!

validation/helpers.py
nienna73: I hadn't even thought to replace all that repeating code, I like this suggestion!

validation/helpers.py
nienna73: Nice! I had been meaning to make this an issue for itwêwina, thanks for jumping on it!

validation/views.py
nienna73: I forgot that I added the auth part to the other views, nice catch

andrewdotn (approved):

validation/helpers.py
eddieantonio: @andrewdotn thanks for fixing this!

I don't really understand what was misconfigured or why, so I'd appreciate an explanation, but that can be whenever :)

Pipfile
eddieantonio: > Based on this thread,
pip install -U virtualenv of all things seemed to have let it install for me, but leaves me even more confused than before.

wat

validation/helpers.py
andrewdotn: Run `openssl s_client -connect sapir.artsrn.ualberta.ca:443 -showcerts < /dev/null` to see a dump of the certs being served up. (*Aside*: an advanced TLS server, such as the one in the go standard library, can dynamically serve up different certificates depending on the capabilities of the client. So this doesn’t work 100% of the time for debugging, but is fine for the predominant case where the server has a static config.)

To see details of individual certs, copy the `-----BEGIN CERTIFICATE-----…-----END CERTIFICATE-----` bit and do `getclip | openssl x509 -text -noout`

The sapir.artsrn.ualberta.ca certificate has `Issuer: C=BE, O=GlobalSign nv-sa, CN=GlobalSign RSA OV SSL CA 2018`, but the intermediate that was being sent (from `intermediate.pem`) had `Subject: C=BE, O=GlobalSign nv-sa, CN=GlobalSign Organization Validation CA - SHA256 - G2`.

It’s likely—but I haven’t checked—that your browser already comes with that specific intermediate certificate, but Python doesn’t, which is why the certificate validates in your browser but wasn’t validating from Python’s `requests.get()`.

The quick and dirty way to find an intermediate certificate is to search the web for sapir’s certificate’s “X509v3 Authority Key Identifier: keyid”, [F8:EF:7F:F2:CD:78:67:A8:DE:6F:8F:24:8D:88:F1:87:03:02:B3:EB](https://google.ca/search?q=F8:EF:7F:F2:CD:78:67:A8:DE:6F:8F:24:8D:88:F1:87:03:02:B3:EB), then look for a PEM Download link on one of the result pages, paste it into a file on the webserver, `apache2ctl -t && apache2ctl restart`, then see if that fixed the problem.

If that fixes things, then that’s usually good enough. There can be cases where it gets really confusing because certificate-authority certificates can be reissued with the same public key but different names/expiration dates/signature algorithms/serial numbers/&c., and some of those variants may be trusted by some clients but not others, which is part of why go supports dynamic certificate selection; but fortunately I’ve never had to deal with that myself.

Pipfile
andrewdotn: wat indeed. I am no longer able to reproduce the issue.

The error that `pipenv` produced was:

    raise VersionConflict(dist, req)
    pkg_resources.VersionConflict: (importlib-metadata 3.4.0 (~/.local/share/virtualenvs/recording-validation-interface-j9JF82f7/lib/python3.7/site-packages), Requirement.parse('importlib-metadata<3,>=0.12; python_version < "3.8"'))

which was confusing because while one `importlib-metadata` constraint was clearly coming from a `pytest` dependency, the `importlib-metadata<3` constraint seemingly came out of nowhere. It turns out it was coming from the implicit `virtualenv` dependency, which has such a constraint:

    $ grep importlib ~/Library/Python/3.9/lib/python/site-packages/virtualenv-20.0.33.dist-info/METADATA
    Requires-Dist: importlib-metadata (<3,>=0.12) ; python_version < "3.8"

So some combination of old/incompatible versions of `pipenv`/`virtualenv` and installing a `Pipfile` for 3.7, using a `pipenv` that may have been running on 3.9, triggered a bug where `virtualenv` had constraints that `pipenv` did not consider and ended up being unsatisfiable.

andrewdotn: To follow up: I have now reproduced the issue. My earlier attempts to do so were failing because, in trying to make a clean repro, I was `pipenv -rm`ing and then forgetting to add the `--dev` flag on install. No `pytest` ⇒ no `VersionConflict`.

With pipenv 2020.11.15 running under Python 3.9.1, and virtualenv version 20.0.33, 20.0.34, 20.0.35, 20.1.0, or 20.2.0 installed, `pipenv install -v black==20.8b1` fails inside a `pipenv shell` with Python 3.7 as the active Python.

Upgrading the *user* install of `virtualenv`, for *Python 3.9*, to 20.2.1 fixed the issue. Presumably later versions of `virtualenv` would too. There must have been some bug where the dependency constraints from the version of `virtualenv` that `pipenv` is running under were leaking into the lock calculation. Python code running inside the pipenv environment (this is *not* where `pipenv` itself runs, because it constructs that environment) only sees `/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/venv`.

eddieantonio: ay dios mio



Thank you for the sluething!

PR #108, opened 2021-01-29
Search

nienna73:
This is a very basic search and doesn't so any sort of fancy matching.

There is an advanced search, which allows users to look for specific analyses or speakers. The process for collecting these results is described in the code.

There are also "filter" buttons on the home page. These filter the shown data by whether or not it's validated and shows all entries by default.

The tests for this branch are pretty basic as I'm not totally sure how to test the full functionality of the advanced search.

### DIFF
--- a/cypress/integration/search.spec.js
+++ b/cypress/integration/search.spec.js
@@ -2,19 +2,19 @@
 // Tests for search and advanced search
 
 describe("Search", () => {
-    it("should display the search results", () => { 
+    it("should search for then display the first word on the main page", () => { 
         cy.visit(Cypress.env("home"))
 
         cy.get('a[name="word-link"]:first')
-            .then((word) => {
-                const w = word[0].id
+            .then((anchorElements) => {
+                const word = anchorElements[0].id
                 cy.get("input[name='query']")
                     .click()
-                    .type(w)
+                    .type(word)
                     .type('{enter}')
 
                 cy.get('h2')
-                    .contains(w)
+                    .contains(word)
                 
                 cy.get(".table")
                     .should("be.visible")
@@ -24,7 +24,7 @@ describe("Search", () => {
                 cy.get(".table")
                     .contains("Recordings")
                 cy.get(".table")
-                    .contains(w)
+                    .contains(word)
             })
     })
 
--- a/cypress/integration/search.spec.js
+++ b/cypress/integration/search.spec.js
@@ -1,0 +1,69 @@
+/// <reference types="cypress" />
+// Tests for search and advanced search
+
+describe("Search", () => {
+    it("should display the search results", () => { 
+        cy.visit(Cypress.env("home"))
+
+        cy.get('a[name="word-link"]:first')
+            .then((word) => {
+                const w = word[0].id
+                cy.get("input[name='query']")
+                    .click()
+                    .type(w)
+                    .type('{enter}')
+
+                cy.get('h2')
+                    .contains(w)
+                
+                cy.get(".table")
+                    .should("be.visible")
+                    .contains("Transcription")
+                cy.get(".table")
+                    .contains("Translation")
+                cy.get(".table")
+                    .contains("Recordings")
+                cy.get(".table")
+                    .contains(w)
+            })
+    })
+
+})
+
+describe("Advanced Search", () => {
+    it("should display the advanced search results", () => { 
+        cy.visit(Cypress.env("home"))
+
+        cy.get('[data-cy="advanced-search-button"]')
+            .click()
+
+        cy.get("input[name='transcription']")
+            .click()
+            .type('kwâskwêpitêw')
+
+        cy.get("input[name='translation']")
+            .click()
+            .type('hook')
+
+        cy.get('[data-cy="advanced-form-submit"]')
+            .click()
+
+        cy.get('h2')
+            .contains("advanced search")
+        
+        cy.get(".table")
+            .should("be.visible")
+            .contains("Transcription")
+        cy.get(".table")
+            .contains("Translation")
+        cy.get(".table")
+            .contains("Recordings")
+        cy.get(".table")
+            .contains("kwâskwêpitêw")
+        cy.get(".table")
+            .contains("hook")
+    })
+
+    // TODO: add tests for analysis matching, validation state matching,
+    // and speaker matchings
+})
--- a/validation/jinja2/validation/advanced_search.html
+++ b/validation/jinja2/validation/advanced_search.html
@@ -50,7 +50,7 @@
 
     <input type="hidden" name="speaker" id="speaker">
 
-    <button onclick="getSelectedSpeakers()" class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
+    <button data-cy="advanced-form-submit" onclick="getSelectedSpeakers()" class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
   </form>
 
   {% endblock content %}
--- a/validation/jinja2/validation/advanced_search.html
+++ b/validation/jinja2/validation/advanced_search.html
@@ -15,7 +15,11 @@
                 selected_speakers += items[i].value+",";
         }
 
-        document.getElementById('speaker').value = selected_speakers;
+        if (selected_speakers.length == 0) {
+            document.getElementById('speaker').value = 'None';
+        } else {
+            document.getElementById('speaker').value = selected_speakers;
+        }
     }
 </script>
 <form method="GET" action="{{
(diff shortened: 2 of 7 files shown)


cypress/integration/search.spec.js
```diff
@@ -0,0 +1,69 @@
+/// <reference types="cypress" />
+// Tests for search and advanced search
+
+describe("Search", () => {
+    it("should display the search results", () => { 
+        cy.visit(Cypress.env("home"))
+
+        cy.get('a[name="word-link"]:first')
+            .then((word) => {
+                const w = word[0].id
```
eddieantonio: Minor note on naming. If I understand this correctly, the thing yielded by `cy.get('a[name="word-link"]:first')` is a `NodeList`. So maybe its name should be something like `anchorElements`? And then the thing you get from it is the `word`. i.e.,

```suggestion
            .then((anchorElements) => {
                const word = anchorElements[0].id
```

cypress/integration/search.spec.js
```diff
@@ -0,0 +1,69 @@
+/// <reference types="cypress" />
+// Tests for search and advanced search
+
+describe("Search", () => {
+    it("should display the search results", () => { 
+        cy.visit(Cypress.env("home"))
+
+        cy.get('a[name="word-link"]:first')
```
eddieantonio, line 8: So this test finds the first word on the page, tries to search for it, then tries to find that word on the search page?  Could you somehow make the intent more clear (comment or function/variable names)?

validation/jinja2/validation/advanced_search.html
```diff
@@ -0,0 +1,56 @@
...
+
+{% block content %}
+<script>
+    function getSelectedSpeakers() {
+        var items = document.getElementsByName('speaker-options')
+        var selected_speakers = "";
+        for (var i=0; i<items.length; i++){
+            if(items[i].type=='checkbox' && items[i].checked==true)
+                selected_speakers += items[i].value+",";
+        }
```
eddieantonio: Suggest using `for..of` loop:

```suggestion
        for (let speaker of items){
            if (speaker.type === 'checkbox' && speaker.checked)
                selectedSpeakers.push(speaker.value);
        }
```

validation/jinja2/validation/advanced_search.html
```diff
@@ -0,0 +1,56 @@
...
+#}
+   
+{% extends 'validation/_base.html' %}
+{% import 'validation/_macros.html' as macros %}
+
+{% block content %}
+<script>
+    function getSelectedSpeakers() {
+        var items = document.getElementsByName('speaker-options')
+        var selected_speakers = "";
```
eddieantonio: Two suggestions:
 - suggest using an array instead
 - consistently use `camelCase` in JavaScript, because everybody else does it ¯\\\_(ツ)\_/¯ 

```suggestion
        var selectedSpeakers = [];
```

validation/jinja2/validation/advanced_search.html
```diff
@@ -0,0 +1,56 @@
...
+        for (var i=0; i<items.length; i++){
+            if(items[i].type=='checkbox' && items[i].checked==true)
+                selected_speakers += items[i].value+",";
+        }
+
+        if (selected_speakers.length == 0) {
+            document.getElementById('speaker').value = 'None';
+        } else {
+            document.getElementById('speaker').value = selected_speakers;
```
eddieantonio:
```suggestion
        if (selectedSpeakers.length === 0) {
            document.getElementById('speaker').value = 'None';
        } else {
            document.getElementById('speaker').value = selectedSpeakers.join(",");
```

validation/views.py
```diff
@@ -59,7 +68,92 @@ def search_phrases(request):
...
+
+
+def advanced_search(request):
+    """
+    The search results for pages.
+    """
+    query = Speaker.objects.all()
+    speakers = []
+    for q in query:
+        speakers.append(q.code)
```
```suggestion
    speakers = [q.code for q in query]
```

validation/views.py
```diff
@@ -59,7 +68,92 @@ def search_phrases(request):
...
+    cree phrases UNION english phrases UNION analysis
+    INTERSECT status
+    INTERSECT speaker
+    """
+    transcription = request.GET.get("transcription")
+    translation = request.GET.get("translation")
+    analysis = request.GET.get("analysis")
+    status = request.GET.get("status")
+    speaker = request.GET.get("speaker")
+    speakers = speaker.strip().split(",")[:-1]
```
eddieantonio: Can replace with [`QueryDict.getlist()`](https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.QueryDict.getlist):

```suggestion
    speakers = request.GET.getlist("speaker-options")
```

(see above JavaScript comments)

validation/jinja2/validation/advanced_search.html
```diff
@@ -0,0 +1,56 @@
+{#
+    # Displays search form to perform an advanced search
+#}
+   
+{% extends 'validation/_base.html' %}
+{% import 'validation/_macros.html' as macros %}
+
+{% block content %}
+<script>
+    function getSelectedSpeakers() {
```
eddieantonio: I have SEVERAL coding suggestions for this method, but I think it may not even be needed at all! Having a bunch of `<input type="checkbox">` with the same name will submit something like this to Django:

```
?transcription=&translation=&status=all&speaker-options=all&speaker-options=MAR&speaker-options=JER&speaker-options=ROS2
```

Which then can be used in accessed in Django as:

```python
>>> request.GET.getlist("speaker-options")
["all", "MAR", "JER", "ROS2"]
```

No JavaScript needed! And as well all know, the best JavaScript is no JavaScript at all! 😇

validation/jinja2/validation/list_phrases.html
```diff
@@ -23,6 +23,15 @@
 {% import 'validation/_macros.html' as macros %}
 
 {% block content %}
+<div>
```
eddieantonio: Is the wrapper `<div>` necessary?

validation/views.py
```diff
@@ -59,7 +68,92 @@ def search_phrases(request):
...
+        elif status == "unvalidated":
+            status_matches = Phrase.objects.filter(validated=False)
+        phrase_and_status_matches = list(
+            set(phrase_matches).intersection(status_matches)
+        )
+    else:
+        phrase_and_status_matches = phrase_matches
+
+    all_matches = []
+    if "all" in speaker or speaker == "None":
```
eddieantonio: With my suggestions, `None` would not appear in the list; it would be empty.

validation/views.py
```diff
@@ -59,7 +68,92 @@ def search_phrases(request):
...
+    if translation != "":
+        english_matches = Phrase.objects.filter(translation__contains=translation)
+    else:
+        english_matches = []
+
+    # if analysis != "":
+    #     analysis_matches = Phrase.objects.filter(analysis__contains=analysis)
+    # else:
+    #     analysis_matches = []
+
```
```suggestion
    # TODO: filter by analysis
```

validation/jinja2/validation/advanced_search.html
```diff
@@ -0,0 +1,56 @@
...
+
+    Speaker:
+    <input type="checkbox" id="all_speakers" name="speaker-options" value="all">
+    <label for="all_spakers">All</label><br>
+        {% for speaker in speakers %}
+        <input type="checkbox" id="{{ speaker }}" name="speaker-options" value="{{ speaker }}">
+        <label for="{{ speaker }}">{{ speaker }}</label><br>
+        {% endfor %}
+
+    <input type="hidden" name="speaker" id="speaker">
```
eddieantonio, line 35: There's probably some fancy Boostrap way of getting the alignment of all of the elements just right, but for this PR, this will do 😅

validation/views.py
```diff
@@ -59,7 +68,92 @@ def search_phrases(request):
     """
     The search results for pages.
     """
-    return HttpResponse(501)
+    query = request.GET.get("query")
+    cree_matches = Phrase.objects.filter(transcription__contains=query)
+    english_matches = Phrase.objects.filter(translation__contains=query)
+    all_matches = list(set().union(cree_matches, english_matches))
```
eddieantonio, line 74: You may want to use `sorted()` so that the order of `all_matches` is consistent between restarts of the Django server. The `set()` built-in is literally random, even Python 3.6 made dictionaries iterate in insertion order.

eddieantonio (changes requested):
I feel like this PR could use some more polish. Feel free to contact me if you have any questions! I have time to Tuesday morning.

validation/jinja2/validation/advanced_search.html
nienna73: Oh this is beautiful! I was not happy with that function either, but I thought it was a necessary evil or sorts. Consider it gone!

validation/jinja2/validation/list_phrases.html
nienna73: Turns out it's not!

eddieantonio (approved):
LGTM!

PR #111, opened 2021-02-03
Translate lemma

nienna73:
If there is no translation for a suggestion in itwêwina, search instead for a translation of the lemma.

Also turns out I was ignoring many possible translations and analyses of a suggestion. Those are now displayed.


validation/helpers.py
```diff
@@ -175,27 +180,65 @@ def get_translations_from_itwewina(word):
...
+    if type(entry["definitions"]) == list and len(entry["definitions"]) > 0:
+        translations = [str(j["text"]) for j in entry["definitions"]]
+
+    translations = "; ".join(translations)
+    analysis = entry["analysis"]
+
+    return translations, analysis
+
+
+def get_lemma(word):
```
andrewdotn: This is very similar to @eddieantonio’s https://gist.github.com/eddieantonio/8f3a678e69f0595dfd1a1e42239e68e0#file-crk_analyze-py-L79-L93 ; if you can wait a bit, we can make a library version to call, though that might not happen until tomorrow

andrewdotn (approved):

validation/helpers.py
nienna73: I can wait for the updated library. There's still quite a bit of refactoring in this branch that will need to be reviewed, but we might as one do one merge with the proper calls to `hfst-optimized-lookup`

eddieantonio: @nienna73 Upgrade to `hfst-optimized-lookup==0.0.10` and then you can do something like this!

```suggestion

fst = TransducerFile("crk-strict-analyzer-for-dictionary.hfstol")

def get_lemma(word):
    try:
        analysis, = fst.lookup_lemma_with_affixes(word)
    except ValueError as err:
        raise ValueError(f"more than one analysis for {word}") from err
    return analysis.lemma
```

nienna73:
Update on this branch: it turns out that once I extract ALL translations from the JSON response and not just the first one, the lemma translation is never used. I've removed it for now, but it will live on in git history if we ever need it again.

andrewdotn (approved):

PR #114, opened 2021-02-09
Docker

nienna73:
It works on Docker! For the most part.

Also all the migration files are in this PR 'cause I guess you're supposed to commit those??


andrewdotn:
The empty `requirements.txt` can be skipped; and looking at the actual migrations now I realize that you may want to [squash them](https://docs.djangoproject.com/en/3.1/topics/migrations/#migration-squashing).

nienna73:
Good catch on that requirements file! I also squashed the migrations and deleted the individual files that the squashed file replaces.

andrewdotn (approved):

Issue #89, opened 2020-09-22
Extract transcriptions for training persephone


aarppe:
Ran the following script which created transcriptions for the recording snippets (with graphemes separed by spaces) and stored them in `/data/av/persephone/SPEAKER/tr/ `(so organized into directories by speakers.

`sqlite3 /data/texts/recval/recval.sqlite3 'SELECT r.id as recording_id, r.speaker_id as speaker, p.transcription as transcription FROM validation_phrase p, validation_recording r WHERE r.phrase_id = p.id;' | gawk '{ tr=""; split($0,s,"|"); n=split(s[3],c,""); for(i=1; i<=n-1; i++) tr=tr sprintf("%s ", c[i]); tr=tr sprintf("%s\n", c[n]); if(s[2]!="") printf "%s", tr > "/data/av/persephone/"s[2]"/tr/"s[1]".tr"; }'`

Issue #88 indicates five erroneously code snippets.

What remains to be done is the extraction of the *.wav files in a similar fashion.

aarppe:
Here's the code that copies the `*.wav` files from `/data/texts/maskwacis-recording-snippets/` to the appropriate speaker-specific directories under `/data/av/persephone/`:

`find /data/texts/maskwacis-recording-snippets/ -path *.wav | gawk 'BEGIN { while(("find /data/av/persephone/ -path *.tr" | getline)!=0) { match($0,"(/data/av/persephone/[^/]+/)tr/(.+)\\.tr$",f); path[f[2]]=f[1]; } } { match($0,"(/data/texts/maskwacis-recording-snippets/)(.+).wav$",f); if(f[2] in path) { system("cp "$0" "path[f[2]]"wav/"); n++; if(n/100==int(n/100)) printf "[%i] ",n > "/dev/stderr"; } } END { printf "==> %i\n", n > "/dev/stderr"; }'`

nienna73:
Both of these extraction processes can be done from within the recording validation project. @eddieantonio implemented the `.wav` extraction and I more recently implemented the transcription portion of this task.

To generate the `.wav` files: `python manage.py importrecordings --wav --skip-db`
To generate the transcription files: `python manage.py writetranscriptions`

Generating the transcription files assumes that you've already imported all the relevant data into the DB and you've already created `.wav` files. These changes can be found in the `write-trans-files` branch and will be merged after investigating a few data oddities.

PR #116, opened 2021-02-10
Write transcription files

nienna73:
The goal of this branch is to write transcription files that can then be used by Persephone and Simple4All as training/test data. I implemented this in the form of a new Django command and added instructions to the Readme. It works well on the small sample of data I have, so hopefully it will work just as well on the 100 000+ entries on Sapir.

This PR relates to issue #89


validation/helpers.py
```diff
@@ -25,7 +25,7 @@
 archive = divvunspell.SpellerArchive(fspath(settings.BASE_DIR / "crk.zhfst"))
 speller = archive.speller()
 
-vowels = ["a", "i", "o", "â", "î", "ô"]
+vowels = ["a", "i", "o", "â", "î", "ô", "e", "ê"]
```
eddieantonio, line 28: lol, yup, we missed `ê`... 🙈

validation/management/commands/writetranscriptions.py
```diff
@@ -0,0 +1,117 @@
...
+    help = "creates transcription files for Persephone and Simple4All"
+
+    def handle(self, *args, **options):
+        # Change this if the raw audio files exist elsewhere!
+        _path = "./audio"
+
+        # Generate this folder by running:
+        # python3 manage.py importrecordings --wav --skip-db
+        audio_dir = Path(_path)
+        for audio_file in os.listdir(audio_dir):
```
eddieantonio:
```suggestion
        for audio_file in audio_dir.iterdir():
```

It also amuses me that [the example in the docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir) just says `: child`, like it's disappointed with the filesystem.

validation/management/commands/writetranscriptions.py
```diff
@@ -0,0 +1,117 @@
...
+            audio_id = audio_file[:-4]
```
eddieantonio: What is the significance of "4"? I assume it's the length of a file extension?

```suggestion
            audio_id = audio_file[:-len(".wav")]
```

If it is just getting rid of the file extension, this is less error-prone:

```suggestion
            audio_id = audio_file.stem
```

See: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem

validation/management/commands/writetranscriptions.py
```diff
@@ -0,0 +1,117 @@
...
+                query = """
+                select speaker_id, transcription from validation_recording as rec, validation_phrase as p
+                where rec.id = ?
+                and rec.phrase_id = p.id;
+                """
+
+                cur.execute(query, (str(audio_id),))
+                result = cur.fetchone()
+                speaker = result[0] if result else None
+                transcription = result[1] if result else None
```
eddieantonio, line 60: If performance becomes an issue, you could perform a [Split Loop refactoring](https://refactoring.com/catalog/splitLoop.html), and open the database one, and fetch ALL of the mappings from {audio_id: (speaker_id, transcription)} from the database in one go.

validation/management/commands/writetranscriptions.py
```diff
@@ -0,0 +1,117 @@
...
+            # Copy the audio file
+            from_dir = audio_dir / audio_file
+            to_dir = audio_dir / speaker / "wav" / audio_file
+            shutil.copyfile(from_dir, to_dir)
+
+            # Treat the transcription for Persephone and save it
+            persephone_filename = audio_id + ".txt"
+            persephone_path = audio_dir / speaker / "label" / persephone_filename
+            persephone_trans = self.create_persephone_transcription(transcription)
+            with open(persephone_path, "w+") as f:
```
eddieantonio:
```suggestion
            with open(persephone_path, "w", encoding="UTF=8") as f:
```

I don't think `"w+"` is appropriate here, because that [opens it for reading too](https://stackoverflow.com/a/30566011/6626414).

I also add `encoding="UTF-8"` all the time, mostly because it's up to the system's default file encoding which is _usually_ UTF-8, but sometimes... isn't 💥 💥 💥

validation/management/commands/writetranscriptions.py
```diff
@@ -0,0 +1,117 @@
...
+                f.write(persephone_trans)
+
+            # Save the transcription for Simple4All
+            s4a_filename = audio_id + ".txt"
+            s4a_path = audio_dir / speaker / "s4a" / s4a_filename
+            with open(s4a_path, "w+") as f:
```
eddieantonio: Same as above :)

eddieantonio (changes requested):
The only suggestion that I think you absolutely should do is either: extract the `-4` magic number, or, even better, remove it entirely. Other than that, this feature looks good! I'm sure you could write an additional command that does the whole documented (🤩🤩🤩) workflow in one command, but that's a PR for another day! 😅

validation/management/commands/writetranscriptions.py
nienna73: image

validation/management/commands/writetranscriptions.py
nienna73: Beautiful.

validation/management/commands/writetranscriptions.py
nienna73: Every time you look at this chunk of code, you find a way to make it more efficient 😂

eddieantonio: I mean, don't optimize until you know it's a problem, but I figure opening and closing the database over 100,000 times is the kind of thing that ends up becoming a performance problem, haha!

eddieantonio (approved):
LGTM!

PR #118, opened 2021-02-11
Audio extraction

nienna73:
This one is going to take a while to review, so I thought I'd try one of those "review as I work on it" things.

These are all the changes made to extract audio info from ELAN files instead of TextGrid files. Also some less than desirable if-statements to find the correct filenames for things.

The print statements help me determine how many folders the system thinks its treated and compare that to the number of sessions that were actually extracted. So they can eventually go, but for now they're helpful.


librecval/extract_phrases.py
```diff
@@ -50,34 +52,47 @@ class MissingMetadataError(RuntimeError):
...
     """
-    All the information you could possible want to know about a recorded
-    snippet.
+    Stores an audio segment extracted from a .eaf file
     """
 
-    session: SessionID
+    translation: str  # in English
+    transcription: str  # in Cree
+    type: str  # "word" or "sentence"
```
andrewdotn: mypy is fussy so this might not be practical, but in *theory* you could

```
# at the very top
from __future__ import annotations
…
from typings import Literal
…
    type: Literal["word"] | Literal["sentence"]
```

librecval/extract_phrases.py
```diff
@@ -50,34 +52,47 @@ class MissingMetadataError(RuntimeError):
...
+class Segment(NamedTuple):
     """
-    All the information you could possible want to know about a recorded
-    snippet.
+    Stores an audio segment extracted from a .eaf file
     """
 
-    session: SessionID
+    translation: str  # in English
+    transcription: str  # in Cree
```
andrewdotn: Two similarly-named variables with comments to help distinguish them? Sometimes you can move the comment into the variable name.

```suggestion
    english_translation: str
    cree_transcription: str
```

librecval/extract_phrases.py
```diff
@@ -86,6 +101,11 @@ def signature(self) -> str:
     def compute_sha256hash(self) -> str:
         """
         Compute a hash that can be used as a ID for this recording.
+        We use the hash instead of including the word in the id for these reasons:
+        - we want people to validate the spelling of the word, so
+        the word itself might change, making the name meaningless
+        - the db doesn't like diacritics very much
```
andrewdotn: The DB *shouldn’t* have a problem with diacritics, as itwêwina’s db happily stores every diacritized wordfrom. Is this something we should look into? Other reasons definitely valid.

librecval/extract_phrases.py
```diff
@@ -133,45 +158,291 @@ def extract_session(self, session_dir: Path):
...
+    sound_bite = audio[start:stop]
+
+    # normalize
+    transcription = normalize(transcription)
+    translation = normalize(translation)
+    sound_bite = sound_bite.normalize(headroom=0.1)
+
+    s = Segment(
+        translation=translation,  # in English
+        transcription=transcription,  # in Cree
```
andrewdotn: Yup, if you gotta add the same comment multiple times when using the var, rename the var.

librecval/extract_phrases.py
```diff
@@ -133,45 +158,291 @@ def extract_session(self, session_dir: Path):
...
+
+@logme.log
+def find_audio_oddities(annotation_path: Path, logger=None) -> Optional[Path]:
+    """
+    Finds the associated audio in Audacity's format.
+    """
+
+    # Get folder name without expected track name
+    i = str(annotation_path).rfind("/")
+    _path = str(annotation_path)[:i]
```
andrewdotn: This looks like what `os.path.dirname()` does; but since it’s a [`Path`](https://docs.python.org/3/library/pathlib.html) object, you can do `folder_name = str(annotation_path.parent)`

librecval/extract_phrases.py
```diff
@@ -133,45 +158,291 @@ def extract_session(self, session_dir: Path):
...
+
+    # the track number is between the last - and the last .
+    j = str(annotation_path).rfind("-")
+    k = str(annotation_path).rfind(".")
+    track = str(annotation_path)[j + 1 : k]
+    track_1 = track.split("_")[0]
```
andrewdotn: I would definitely use a regular expression here, but not everyone would necessarily find that cleaner.

    # the track number is between the last - and the last .
    match = re.match(r'.*-([^-]+)\.[^-.]+', str(annotation_path))
    if not match:
        raise Error("could not identify track number)
    track_number = match.group(1)

librecval/extract_phrases.py
```diff
@@ -215,153 +486,25 @@ def find_audio_from_audition_format(
...
     2
-    >>> get_mic_id('2015-05-11am-03.TextGrid')
+    >>> get_mic_id('2015-05-11am-03.eaf')
     3
-    >>> get_mic_id('2016-02-24am-Track 2_001.TextGrid')
+    >>> get_mic_id('2016-02-24am-Track 2_001.eaf')
     2
-    >>> get_mic_id('Track 4_001.TextGrid')
+    >>> get_mic_id('Track 4_001.eaf')
     4
```
andrewdotn, line 519: Doctests are great!

andrewdotn (commented):
It’s often really useful to have lots of statements that print info; using the logger interface can be fidgety to set up but then lets you turn the log statements on and off in different parts of the code. I’m not familiar with the `logme` library here so I don’t know how easy it is to get that working.

librecval/extract_phrases.py
eddieantonio: Sapir's filesystem and BACKUP UTILITY have issues with diacritics 😩 

So this is why I created hashes. Basically, so there'd be no issues when backing up the audio.

nienna73: Also! When I didn't use the hash, I had some nasty URL issues

librecval/extract_phrases.py
nienna73: I like this suggestion, but I'm hesitant to incorporate something that might not work. I do think this is clearer, but when the extraction process takes 4+ hours to run, I really don't want it stopping on a typing issue. Do you think the suggestion is safe enough for this scenario?

librecval/extract_phrases.py
nienna73: I actually don't mind using regex, but it can be harder to understand at a glance 🤔 @eddieantonio which do you prefer?

nienna73: There are also places later on where having the k index is useful:
Line 420: `track_8 = str(annotation_path)[i + 1 : k]` to get just the filename (no extension)
Line 437: `track_13 = str(annotation_path)[i + 1 : k]` again, to get just the filename (no extension)

I'm sure there's a Pathlib way of doing this that would be cleaner, so please teach me!

eddieantonio: Regex is fine, but I like using verbose patterns like the ones found in [`get_mic_id()`](https://github.com/UAlbertaALTLab/recording-validation-interface/blob/6d398d4ccd0c7294cc90965b1aa8277a29432702/librecval/extract_phrases.py#L508-L547), because regex can become difficult to read and maintain!

To get the filename with no extension with Path, use [`annotation_path.stem`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem)

librecval/extract_phrases.py
eddieantonio: Since this is a change that affects static typing, you could figure out if `mypy` is upset with this _before running the extraction process_. So, yes, it's safe!

eddieantonio: Since "word" and "sentence" are values that can come from code (they don't come from any sort of user input), this is safe!

Here are some further suggestions:

```python
# add this to the imports
from typings import Literal

# add this to the globals/constants
WordOrSentence = Literal["word", "sentence"]

# change the named tuple:
class Segment(NamedTuple):
    # ...
    type: WordOrSentence

# add typing to the definition of extract_data():
def extract_data(
    _file, _type: WordOrSentence, snippet, audio, speaker, session_id, english_tier, comment_tier
) -> tuple:
```

**EDIT**: use correct syntax for `Literal` (see: https://mypy.readthedocs.io/en/stable/literal_types.html#parameterizing-literals)

librecval/extract_phrases.py
```diff
@@ -344,20 +345,28 @@ def find_audio_oddities(annotation_path: Path, logger=None) -> Optional[Path]:
...
+    )
+    if match:
+        track = match.group(1)
+        track_1 = track.split("_")[0]
 
-    # try 1: the .wav file is in a subfolder, but it has 'Track number' in it
-    dirs = list(glob.glob(_path + "/**/" + track + "*.wav", recursive=True))
-    sound_file = Path(dirs[0]) if len(dirs) > 0 and Path(dirs[0]).exists() else None
+        # try 1: the .wav file is in a subfolder, but it has 'Track number' in it
+        dirs = list(glob.glob(_path + "/**/" + track + "*.wav", recursive=True))
```
andrewdotn:
```suggestion
        dirs = list(_path.glob("/**/" + track + "*.wav"))
```

eddieantonio: PATHS FOR THE PATHLIB CROWN! F-STRINGS FOR THE F-STRING HORDE

```suggestion
        dirs = list(_path.glob(f"**/{track}/*.wav"))
```

https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob

librecval/extract_phrases.py
nienna73: Thanks for the suggestions and clarification!

librecval/extract_phrases.py
nienna73: Do these changes keep the "recursive" property? I remember it being a pain to figure out that I needed that argument

andrewdotn: Yup, the link says,

> The “**” pattern means “this directory and all subdirectories, recursively”. In other words, it enables recursive globbing

nienna73: Oh perfect, thank you!
I was using `glob.glob` in a lot of places...

librecval/extract_phrases.py
```diff
@@ -28,12 +28,15 @@
 from os import fspath
 from pathlib import Path
 from typing import Dict, NamedTuple, Optional
+import glob
```
eddieantonio: Unused import:

```suggestion
```

(all uses globs are using `pathlib`, not `glob`)

validation/models.py
```diff
@@ -345,7 +346,14 @@ class Recording(models.Model):
     # TODO: determine this automatically during import process
     quality = models.CharField(
         help_text="Is the recording clean? Is it suitable to use publicly?",
-        **arguments_for_choices(QUALITY_CHOICES),
+        # **arguments_for_choices(QUALITY_CHOICES),
```
eddieantonio: How come this is commented out? Can this be deleted entirely?

Does `QUALITY_CHOICES` need to exist at all in this PR?

eddieantonio (approved):
LGTM! I have minor comments, but it basically amounts to deleting a few lines of code 😅

validation/models.py
nienna73: I didn't actually write `QUALITY_CHOICES` but I can delete it

PR #127, opened 2021-03-01
Pagination on Search

### LINKED ISSUE #126: Fix Pagination
Pagination does not currently work from either search results view (advanced search and basic search). Clicking "Next" brings up and error about empty query strings, since we need the search terms to continue.

We need to either pass the query to the pagination, or make it so it doesn't always need a query.

nienna73:
fix #126

Pagination was broken for both basic and advanced search. I now pass the query parameters through the view so the paginator has access to them when loading the next page of results.

I also changed how many results show per page from 30 --> 5, though 5 might be too few results. I'm open to suggestions.

### DIFF
--- a/validation/views.py
+++ b/validation/views.py
@@ -86,10 +86,13 @@ def search_phrases(request):
     english_matches = Phrase.objects.filter(translation__contains=query)
     all_matches = list(set().union(cree_matches, english_matches))
     all_matches.sort(key=lambda phrase: phrase.transcription)
-    paginator = Paginator(all_matches, 30)
+
+    query_term = f"&query={query}"
+
+    paginator = Paginator(all_matches, 5)
     page_no = request.GET.get("page", 1)
     phrases = paginator.get_page(page_no)
-    context = dict(phrases=phrases, search_term=query)
+    context = dict(phrases=phrases, search_term=query, query=query_term)
     return render(request, "validation/search.html", context)
 
 
@@ -159,10 +162,16 @@ def advanced_search_results(request):
 
     all_matches.sort(key=lambda phrase: phrase.transcription)
 
-    paginator = Paginator(all_matches, 30)
+    query = f"&transcription={transcription}&translation={translation}&analysis={analysis}&status={status}"
+    for speaker in speakers:
+        query += f"&speaker-options={speaker}"
+
+    query += "&speaker="
+
+    paginator = Paginator(all_matches, 5)
     page_no = request.GET.get("page", 1)
     phrases = paginator.get_page(page_no)
-    context = dict(phrases=phrases, search_term="advanced search")
+    context = dict(phrases=phrases, search_term="advanced search", query=query)
     return render(request, "validation/search.html", context)
 
 
--- a/validation/views.py
+++ b/validation/views.py
@@ -371,4 +371,4 @@ def register(request):
 
 def encode_query_with_page(query, page):
     query["page"] = page
-    return f"?{query.urlencode()}" if query else ""
+    return f"?{query.urlencode()}"
--- a/validation/views.py
+++ b/validation/views.py
@@ -366,4 +366,4 @@ def register(request):
 
 def encode_query_with_page(query, page):
     query["page"] = page
-    return query.urlencode()
+    return f"?{query.urlencode()}" if query else ""
--- a/validation/views.py
+++ b/validation/views.py
@@ -88,7 +88,8 @@ def search_phrases(request):
     all_matches = list(set().union(cree_matches, english_matches))
     all_matches.sort(key=lambda phrase: phrase.transcription)
 
-    query_term = QueryDict(f"query={query}", mutable=True)
+    query_term = QueryDict("", mutable=True)
+    query_term.update({"query": query})
 
     paginator = Paginator(all_matches, 5)
     page_no = request.GET.get("page", 1)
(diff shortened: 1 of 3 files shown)


validation/views.py
```diff
@@ -86,10 +86,13 @@ def search_phrases(request):
     english_matches = Phrase.objects.filter(translation__contains=query)
     all_matches = list(set().union(cree_matches, english_matches))
     all_matches.sort(key=lambda phrase: phrase.transcription)
-    paginator = Paginator(all_matches, 30)
+
+    query_term = f"&query={query}"
```
eddieantonio: How does quoting/escaping work here? Does `query` need to be urlencoded? Or is it _already_ URL-encoded?

validation/views.py
```diff
@@ -159,10 +162,16 @@ def advanced_search_results(request):
 
     all_matches.sort(key=lambda phrase: phrase.transcription)
 
-    paginator = Paginator(all_matches, 30)
+    query = f"&transcription={transcription}&translation={translation}&analysis={analysis}&status={status}"
+    for speaker in speakers:
+        query += f"&speaker-options={speaker}"
+
+    query += "&speaker="
```
eddieantonio: Same comment here: what is the deal with quoting? Is user-input trusted? Is it sanitized before it gets here?

Perhaps consider using [urllib.parse.urlencode](https://docs.python.org/3.7/library/urllib.parse.html#urllib.parse.urlencode)?

```python
>>> from urllib.parse import urlencode
>>> urlencode({"transcrption": "atim", "translation": "dog", "analysis": "atim+N+A+Sg", "speaker": "MAR"})
'transcrption=atim&translation=dog&analysis=atim%2BN%2BA%2BSg&speaker=MAR'
```

(notice what happened to the `analysis`!)

eddieantonio (commented):
I have some concerns about URL encoding!

andrewdotn (commented):
For query strings, it’s much safer to use [QueryDict](https://docs.djangoproject.com/en/3.1/ref/request-response/#querydict-objects) objects. They’re a lot like normal dicts except they have a `urlencode()` method for when they need to go into HTML at the end.

That way things won’t break mysteriously if there’s a stray `&` or `=` character in a translation or search string, and it also provides some protection against more malicious actions too.

Since this is jinja, I *think* you can get a function into the template by just passing it into the context?

    def encode_query_with_page(query, page):
      query['page'] = page
      return query.urlencode()

    context = dict(…, encode_query_with_page=encode_query_with_page)

and then in the template

    href="{{ encode_query_with_page(query, page.next_page_number()) }}"

This can likely be simplified a bit, but I wanted to at least give some pointers when asking you to do more work here.

validation/views.py
nienna73: Ah! Oh no! I will fix this.

eddieantonio:
> For query strings, it’s much safer to use [QueryDict](https://docs.djangoproject.com/en/3.1/ref/request-response/#querydict-objects) objects. They’re a lot like normal dicts except they have a `urlencode()` method for when they need to go into HTML at the end.
> 
> That way things won’t break mysteriously if there’s a stray `&` or `=` character in a translation or search string, and it also provides some protection against more malicious actions too.
> 
> Since this is jinja, I _think_ you can get a function into the template by just passing it into the context?
> 
> ```
> def encode_query_with_page(query, page):
>   query['page'] = page
>   return query.urlencode()
> 
> context = dict(…, encode_query_with_page=encode_query_with_page)
> ```
> 
> and then in the template
> 
> ```
> href="{{ encode_query_with_page(query, page.next_page_number()) }}"
> ```
> 
> This can likely be simplified a bit, but I wanted to at least give some pointers when asking you to do more work here.

I forgot about `QueryDict`. Very nice solution to this!

nienna73:
Using `QueryDict` and `urlencode` returns a string, such as "search/query=apa&page=2", but the lack of '?' means it returns an invalid url for the app. Should I manually add the '?' in `_macros`, or is there a better way of doing this?

andrewdotn:
Yup, you can add the `?` in `_macros`, as `https://…/blah?` is (always?) functionally equivalent to `https://…/blah`. But if you want to address the aesthetic issue, and already have a function encoding the query, that function can do something like `return ("?" + encoded query) if encoded_query else ""`

nienna73:
I would think it should work with or without the `?`, but Django disagrees. Or I've set up Django to disagree, I'm not too sure which 🤔

validation/views.py
```diff
@@ -87,12 +88,17 @@ def search_phrases(request):
     all_matches = list(set().union(cree_matches, english_matches))
     all_matches.sort(key=lambda phrase: phrase.transcription)
 
-    query_term = f"&query={query}"
+    query_term = QueryDict(f"query={query}", mutable=True)
```
```suggestion
    query_term = QueryDict('', mutable=True)
    query_term.update({"query": query})
```

validation/views.py
```diff
@@ -162,16 +168,23 @@ def advanced_search_results(request):
...
     all_matches.sort(key=lambda phrase: phrase.transcription)
 
-    query = f"&transcription={transcription}&translation={translation}&analysis={analysis}&status={status}"
+    query = f"transcription={transcription}&translation={translation}&analysis={analysis}&status={status}"
     for speaker in speakers:
         query += f"&speaker-options={speaker}"
 
     query += "&speaker="
 
+    query = QueryDict(query, mutable=True)
```
```suggestion
    query = QueryDict('', mutable=True)
    query.update({"transcription": transcription, "translation": translation, "analysis": analysis, "status": status})
    for speaker in speakers:
        query.appendlist('speaker-options', speaker)
```

andrewdotn (commented):
Ok, `QueryDict` has a messier API than I expected. I was going to say, its complexity may not be worth it unless you’re using multi-valued options, but then I see that you are using multi-valued options—

I’ve tried to show how to use it.

Basically, any time you have some variable `foo` whose contents may have come from a user, you never want to do something like `f"blah blah {foo} bar"` that will directly include the value of `foo` into a web page or URL. It needs to go through some library class/function like `QueryDict` or `urlencode` so that characters get escaped properly.

(In a django/jinja template, the template library does this automatically for you behind the scenes, so `blah blah {{foo}} blah` *is* generally safe there.)

validation/views.py
```diff
@@ -340,3 +367,8 @@ def register(request):
 
 
 # TODO: Speaker bio page like https://ojibwe.lib.umn.edu/about/voices
+
+
+def encode_query_with_page(query, page):
+    query["page"] = page
+    return f"?{query.urlencode()}" if query else ""
```
eddieantonio: I don't know why but this chunk oof code is massively confusing to me.

andrewdotn (approved):

PR #131, opened 2021-03-02
Auto validation

nienna73:
Auto validation works and is set up as its own Django management command. After following all the other steps, run `./manage.py autoval` to auto-validate the entries.

Right now, auto validation updates the transcription, analysis, status, and validated fields of the Phrase object. I am also able to say if the auto validation suggestion comes from CW or MD--should I use this to update the `origin` field of the Phrase, or not yet?

I refactored the Phrase model to have a `transcription` and a `field_transcription`. When users make edits/suggestions to the spellings of things, it will change the `transcription` field; the `field_transcription` should never change. I may have over-refactored this variable when it comes to the tests--please check those carefully!
When a phrase gets added to the DB, `transcription` is set to `field_transcription`, so we should be able to always access `transcription` when looking at or modifying things.

The new status types are also included in this PR. When a phrase is first added to the DB, it's given the status `new`. If it gets auto-standardized, it gets the status `auto-standardized`. A Phrase will be: `linked` when the transcription and translation have been confirmed to match; `standardized` when it's been viewed by a linguist and given a lemma/analysis; and `validated` when its been standardized and has at least one good recording. These terms are still being defined as we continue to think of how users will interact with the site.


validation/helpers.py
```diff
@@ -206,9 +206,45 @@ def get_distance_with_translations(word):
...
+    # MED = 0 or an exact match for spelling
+    # Returns none if no matches found, or if multiple matches found
+
+    match = None
+    for suggestion in suggestions:
+        if suggestions[suggestion]["med"] == 0 or suggestion == word:
+            if match is None:
+                match = suggestions[suggestion]
+            else:
+                return None
```
eddieantonio, line 231: Ah, so this is in the case of a duplicate?

aarppe:
Autovalidation should solely be done on the basis of whether a field transcripted word can be 1) accepted as is, or be 2) provided a single string with MED=0, using Divvunspell. So a match with old MD content is not sufficient for now, since the MD orthography is underspecified wrt vowel length and aspiration.

eddieantonio:
> Right now, auto validation updates the transcription, analysis, status, and validated fields of the Phrase object. I am also able to say if the auto validation suggestion comes from CW or MD--should I use this to update the origin field of the Phrase, or not yet?

As far as I understand, all words in the Maskwacîs content are either from MD or they are 🆕  words, freshly-ellicited. @aarppe, please confirm.

aarppe:
Yes. And unless the MD words have been explicitly linked with a CW entry (based on Katie's work), those other MD words cannot be considered standardized.

It will become a separate task to link remaining MD entries with CW entries, which may require the addition of those words into the LEXC source for the crk FST.

nienna73:
Okay, I won't update the `origin` field during the auto-validation process.

validation/helpers.py
```diff
@@ -206,9 +206,45 @@ def get_distance_with_translations(word):
...
             "len": len(matches) if len(matches) == 1 else len(matches) + 1,
         }
 
     return suggestions
+
+
+def perfect_match(word, suggestions):
+    # Checks for exactly one entry with either
+    # MED = 0 or an exact match for spelling
+    # Returns none if no matches found, or if multiple matches found
```
eddieantonio: Nitpick: convert comments like these to docstrings so that automatic documentation discovery tools can easily pull it out.

```suggestion
    """
    Checks for exactly one entry with either
    MED = 0 or an exact match for spelling
    Returns none if no matches found, or if multiple matches found
    """
```

validation/management/commands/autoval.py
```diff
@@ -0,0 +1,55 @@
...
+
+                # only save the analysis if there is exactly one
+                if exactly_one_analysis(_match):
+                    phrase.transcription = _match["transcription"]
+                    phrase.analysis = _match["matches"][0]["analysis"]
+                    phrase.status = "auto-validated"
+                    phrase.validated = True
+                    phrase.save()
+
+        return
```
eddieantonio:
```suggestion
```

eddieantonio (commented):
I will continue reviewing tomorrow, but just a few comments for today.

validation/helpers.py
nienna73: Yes. It's not the most elegant way of doing it, but it does fail fast in the case of failure.

validation/management/commands/autoval.py
```diff
@@ -0,0 +1,53 @@
...
+        for phrase in tqdm(phrases):
+            if not phrase.validated:
+                segment_name = phrase.field_transcription
+                suggestions = get_distance_with_translations(segment_name)
+                _match = perfect_match(segment_name, suggestions)
+
+                # only save the analysis if there is exactly one
+                if exactly_one_analysis(_match):
+                    phrase.transcription = _match["transcription"]
+                    phrase.analysis = _match["matches"][0]["analysis"]
```
eddieantonio: `match` is an acceptable name (i.e., it's not a [Python built-in](https://docs.python.org/3/library/functions.html#built-in-funcs))!

I also like how readable this code is! Like, oh, if the phrase is not validated, find a perfect match, make sure there's only one possible analysis of it, then autovalidate it!

```suggestion
                match = perfect_match(segment_name, suggestions)

                # only save the analysis if there is exactly one
                if exactly_one_analysis(match):
                    phrase.transcription = match["transcription"]
                    phrase.analysis = match["matches"][0]["analysis"]
```

validation/models.py
```diff
@@ -67,18 +67,40 @@ class Phrase(models.Model):
...
+
     MASKWACÎS_DICTIONARY = "MD"
     NEW_WORD = "new"
     ORIGIN_CHOICES = (
         (MASKWACÎS_DICTIONARY, "Maskwacîs Dictionary"),
         (NEW_WORD, "New word"),
     )
 
+    field_transcription = models.CharField(
+        help_text="The field transciption of the Cree phrase.",
```
eddieantonio: What does "field transcription" mean?

validation/models.py
```diff
@@ -151,8 +180,10 @@ def clean(self):
         """
         Cleans the text fields.
         """
-        self.transcription = normalize_sro(self.transcription)
-        assert self.ALLOWED_TRANSCRIPTION_CHARACTERS.issuperset(self.transcription)
+        self.field_transcription = normalize_sro(self.field_transcription)
```
eddieantonio: Should the _field transcription_ be automatically normalized? You may want to automatically convert to unicode NFC normalization form to prevernt future headaches (`unicodedata.normalize("NFC", self.field_transcription)`); but the field transcription is the "original" transcription, right? So maybe this should not be cleaned; left as is.

validation/tests/test_recordings_search.py
```diff
@@ -59,7 +59,7 @@ def test_search_recordings(client, bake_recording):
     assert isinstance(recordings, list)
     assert len(recordings) == 1
     recording = recordings[0]
-    assert recording.get("wordform") == phrase.transcription
+    assert recording.get("wordform") == phrase.field_transcription
```
eddieantonio: Hmm... I'm unsure about this one. So, this is when itwêwina asks the speech-db for recordings for, e.g., "wâpamêw". Although this is (maybe?) a breaking change, I think we want it to search for the _validated_ transcription rather than the original transcription!

```suggestion
    assert recording.get("wordform") == phrase.transcription
```

Basically: when itwêwina wants a recording for a word (and you can assume itwêwina has "validated" wordforms), what should it match in the database? This same comment goes for the rest of the tests in this file.

validation/views.py
```diff
@@ -267,7 +273,7 @@ def make_absolute_uri_for_speaker(code: str) -> str:
 
         recordings.extend(
             {
-                "wordform": rec.phrase.transcription,
+                "wordform": rec.phrase.field_transcription,
```
eddieantonio: Again, maybe this should be `transcription` instead?

eddieantonio (commented):
i am confusion by `field_translation` in the recording API. Feel free to chat with me to clarify what the recording API is actually intended to do!

eddieantonio (approved):
LGTM!

PR #137, opened 2021-03-04
Default user group

nienna73:
New users are added to the Community group when they register. If the group doesn't exist, it will be created. Users who are not community members will need to request access to be added to the Linguist group.


eddieantonio (approved):
LGTM!

Nice and simple and doesn't require a migration!

PR #138, opened 2021-03-04
Use Cards on the home page

nienna73:
## In this PR
This is the beginning of the UI overhaul PRs. These changes are included:
* use cards on the homepage instead of a table
* the user can validate a transcription from the home page


validation/helpers.py
```diff
@@ -11,6 +11,8 @@
 from django.conf import settings
 from django.utils.http import urlencode
 
+import validation
```
eddieantonio: Oh lawd, why does this need to be imported?

validation/jinja2/validation/search.html
```diff
@@ -25,49 +25,44 @@
 
 {% block content %}
 <h2> Results for <q>{{ search_term }}</q> </h2>
+    {%  for phrase in phrases %}
+    <div class="card segment-card">
```
eddieantonio: [`<article>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) may be more appropriate here:

```suggestion
    <article class="card segment-card">
```

Each card is its own self-contained entry, correct?

validation/jinja2/validation/search.html
```diff
@@ -25,49 +25,44 @@
...
-          <li>
-            <div class="rec-height">
-              {{ macros.speaker(recording) }}
-            </div>
-          </li>
+      </div>
+    {% endif %}
+      <div class="card-body">
+        <h3>{{ phrase.transcription }}</h3>
+        <h5>Translation: {{ phrase.translation }}</h5>
```
eddieantonio: Super weird to jump from `<h3>` to `<h5>`. I'm not sure what the intention is here, but you could use `<p>` with a class like `.card__translation` (using [BEM notation](http://getbem.com/naming/#:~:text=Element,-Parts)).

validation/jinja2/validation/search.html
```diff
@@ -25,49 +25,44 @@
...
+            <th>Speaker</th>
 
-        {% endfor %}{# for recording in phrase.recordings #}
-        </ul>
-      </td>
-    </tr>
-    {% endfor %}{# for phrase in page.phrases#}
-  </table>
+            {% for recording in phrase.recordings %}
+                <tr>
```
eddieantonio: Idk what happened to the indentation here but it confused me!

```suggestion
            <tr>
```

validation/jinja2/validation/search.html
```diff
@@ -25,49 +25,44 @@
...
+                <td>
+                    {{ macros.speaker(recording) }}
+                </td>
+                </tr>
+
+            {% endfor %}{# for recording in phrase.recordings #}
+        </table>
+        <a href={{ url('validation:segment_detail', phrase.id) }} class="button-success right">More Options</a>
+      </div>
 </div>
```
eddieantonio:
```suggestion
</article>
```

I think...?

validation/static/index.js
```diff
@@ -0,0 +1,37 @@
+"use strict";
+
+//    translation-judgement-accuracy-yes
+document.addEventListener('DOMContentLoaded', () => {
+    console.log("hello")
+    console.log(csrftoken)
```
eddieantonio:
```suggestion
```

validation/static/index.js
```diff
@@ -0,0 +1,37 @@
...
+    for (let judgement of ["yes", "no"]) {
+        for (let button of document.querySelectorAll(`.translation-judgement-accuracy-${judgement}`)) {
+            button.addEventListener("click", async (e) => {
+                console.log("You clicked on", e)
+                console.log(e.target.dataset.phraseId)
```
eddieantonio:
```suggestion
```

validation/static/index.js
```diff
@@ -0,0 +1,37 @@
...
+                    body: JSON.stringify({judgement})
+                })
+                await response.json().then((r) => {
+                    console.log(r['status'])
+                    if (r['status'] === 'ok') {
+                        if (judgement === 'yes') {
+                            button.setAttribute("class", "button-success-solid translation-judgement-accuracy-yes")
+
+                        } else if (judgement === 'no') {
+                            button.setAttribute("class", "button-fail-solid translation-judgement-accuracy-yes")
```
eddieantonio: ???

```suggestion
                            button.setAttribute("class", "button-fail-solid translation-judgement-accuracy-no")
```

validation/static/index.js
```diff
@@ -0,0 +1,37 @@
...
+                        }
+                    }
+                })
+            })
```
eddieantonio: I tried refactoring this because I was confused by the use of both `Promise` and `async`/`await` syntax in the same place.

```suggestion
                let r = await response.json()

                if (r.status !== 'ok') {
                    return
                }

                let buttonStatus;

                if (judgement === 'yes') {
                    buttonStatus = 'success'
                } else if (judgement === 'no') {
                    buttonStatus = 'fail'
                }

                button.setAttribute("class", `button-${buttonStatus}-solid translation-judgement-accuracy-yes`)
```

validation/static/styles.css
```diff
@@ -135,7 +135,23 @@ li.speaker_code {
   transition: color .15s ease-in-out, background-color .15s ease-in-out, border-color .15s ease-in-out, box-shadow .15s ease-in-out;
 }
 
-.button-success:hover {
+.button-success-solid {
```
eddieantonio: A few notes on refactoring CSS:

I have found the philosophy of [BEM](http://getbem.com/naming/) and [Tailwind CSS](https://tailwindcss.com/) to be quite helpful in DRYing up my rules.

So I would make a `.button` class that contains all common code. And then I'd make classes called `.button--success` and `.button--fail` that would override the colours.

I'm not immediately sure if there's a distinction between `.button--solid` and... `.button--not-solid`, but either way, I assume _most_ of those style changes to be orthogonal to `.button--success`/`.button--fail` (e.g., there's little interaction).

Usage would be:

```
<button type="submit" class="button button--solid button--success">...</button>
<a href="#" class="button button--solid button--success">...</a>
```

etc.

validation/views.py
```diff
@@ -380,6 +382,23 @@ def register(request):
 # TODO: Speaker bio page like https://ojibwe.lib.umn.edu/about/voices
 
 
+@require_http_methods(["POST"])
+def record_translation_judgement(request, phrase_id):
+    # TODO: check that user is logged in
```
eddieantonio, line 389: Will `@login_required` help here?

https://docs.djangoproject.com/en/3.1/topics/auth/default/#the-login-required-decorator

validation/views.py
```diff
@@ -380,6 +382,23 @@ def register(request):
...
+    phrase = get_object_or_404(Phrase, id=phrase_id)
+    print(phrase)
```
```suggestion
```

validation/views.py
```diff
@@ -380,6 +382,23 @@ def register(request):
...
+    judgement = json.loads(request.body)
+
+    if judgement["judgement"] == "yes":
+        phrase.validated = True
+    elif judgement["judgement"] == "no":
+        phrase.validated = False
```
eddieantonio, line 396:
```suggestion
    if judgement["judgement"] == "yes":
        phrase.validated = True
    elif judgement["judgement"] == "no":
        phrase.validated = False
    else:
        return HttpResponseBadRequest()
```

Defensive programming!

validation/views.py
```diff
@@ -380,6 +382,23 @@ def register(request):
...
+
+    phrase.save()
+    print("got request")
```
```suggestion
```

eddieantonio (commented):
I like the cards idea a lot better than a giant, intimidating table! 😅

validation/helpers.py
nienna73: It doesn't! Good catch

validation/jinja2/validation/search.html
nienna73: Oof, me neither. I'll fix all the indentation.

validation/static/index.js
nienna73: 🙈

validation/views.py
nienna73: That does look useful, but I'll address this in its own PR.

validation/static/styles.css
nienna73: `button--solid` is to indicate the current state of a segment, so it's somewhere between "selected" and "unselected". That is, the background colour is just at 80% opacity.

validation/jinja2/validation/logout.html
```diff
@@ -2,7 +2,7 @@
 
 
 {% block content %}
-<div class="logout-message">
+<div class="logout__message">
```
eddieantonio: This doesn't seem to be the "element" of a `logout` block, so it's fine to have a name just like `logout-message`.

eddieantonio (approved):
LGTM!

PR #139, opened 2021-03-04
User permissions

nienna73:
## In this PR:
* only linguists can view more options for each segment
* linguists can view auto-validated phrases, community members can not
* styles for the cards change on button press (status change)
* the filter buttons on the top of the home screen represent the new Phrase statuses
* the current username is saved in the Phrase history when a Phrase status is updated

**Notes:**
- PR #138 should be approved and merged before this one
- The behaviours and permissions on the search results pages do not yet match the behaviours and permissions on the home page


validation/views.py
```diff
@@ -49,19 +49,29 @@ def index(request):
...
     The home page.
     """
+    is_linguist = user_is_linguist(request.user)
+
     all_class = "button button--success filter__button"
-    validated_class = "button button--success filter__button"
-    unvalidated_class = "button button--success filter__button"
+    new_class = "button button--success filter__button"
+    linked_class = "button button--success filter__button"
+    auto_validated_class = "button button--success filter__button"
```
eddieantonio, line 57: I should have mentioned this before, but describing the requisite classes in the `views.py` file is maybe not the best place to put them — all class based stuff should really be in the templates.

But we'll save that for a future PR! 😅

validation/static/index.js
```diff
@@ -25,8 +25,36 @@ document.addEventListener('DOMContentLoaded', () => {
 
                 if (judgement === 'yes') {
                     button.setAttribute("class", "button button--success-solid translation-judgement-accuracy-yes")
+                    
+                    const noButtons =  document.getElementsByClassName("translation-judgement-accuracy-no");
+                    for (let b of noButtons) {
+                        if (b.dataset.phraseId === phraseId) {
+                            b.setAttribute("class", "button button--fail translation-judgement-accuracy-no")
```
eddieantonio: You may want to look into [`b.classList`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#examples) instead of using `b.setAttribute()`, e.g.,

```suggestion
                            b.classList.remove("button--success")
                            b.classList.add("button--fail")
```

validation/static/index.js
```diff
@@ -25,8 +25,36 @@ document.addEventListener('DOMContentLoaded', () => {
...
+                        }
+                    }
+
+                    const headers = document.getElementsByClassName("card__header")
+                    for (let h of headers) {
+                        if (h.dataset.phraseId === phraseId) {
+                            h.setAttribute("class", "card__header card__header--green")
```
eddieantonio: Again, with `element.classList`:

```suggestion
                            h.classList.add("card__header--green")
```

validation/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 2.2.19 on 2021-03-02 22:30
+# Generated by Django 2.2.19 on 2021-03-04 20:36
```
eddieantonio, line 1: Weird! Why did this migration change its date but nothing else?

validation/jinja2/validation/list_phrases.html
```diff
@@ -29,32 +29,38 @@
...
+        {% endif %} {# if is_linguist #}
     </div>
   </form>
 </div>
 
     {%  for phrase in phrases %}
     <div class="card shadow-sm">
     {% if phrase.validated %}
-      <div class="card__header card__header--green">
+      <div class="card__header card__header--green" data-phrase-id={{ phrase.id }}>
```
eddieantonio, line 44: I like the use of these `data-*` attributes!

eddieantonio (approved):
LGTM! There's room for improvement, but it's fine without it.

validation/migrations/0001_initial.py
nienna73: I...may have deleted it?

validation/static/index.js
nienna73: I like this better--fixed all throughout this file!

validation/views.py
nienna73: I completely agree, I just didn't know how to do it. I think it would take a tad more time than we have now so I'll make an issue for it.

eddieantonio (approved):
LGTM

PR #140, opened 2021-03-05
Persephone

nienna73:
## What's in this PR:
This is mainly a refactor of the code that writes transcription files and relocates the audio files so Persephone and Simple4All can use them for training. It uses the `field_transcription` when generating the transcriptions for all the files, but then it uses the `transcription` when generating just the auto-generated transcriptions.

It stills feels a bit messy to me, but it's a little better at least.

I'm unsure if the path "speaker/auto_val/label" will actually work for training, but the files are easy to move once generated.

### DIFF
--- a/validation/management/commands/writetranscriptions.py
+++ b/validation/management/commands/writetranscriptions.py
@@ -43,31 +43,32 @@ class Command(BaseCommand):
         # python3 manage.py importrecordings --wav --skip-db
         audio_dir = Path(_path)
 
-        query = """
-            select speaker_id, field_transcription from validation_recording as rec, validation_phrase as p
-            where rec.id = ?
-            and rec.phrase_id = p.id;
-            """
-
         print("Writing transcriptions for all phrases")
-        self.write_transcriptions(audio_dir, query)
-
-        query = """
-            select speaker_id, transcription from validation_recording as rec, validation_phrase as p
-            where rec.id = ?
-            and rec.phrase_id = p.id
-            and p.status = "auto-validated";
-            """
+        self.write_transcriptions(audio_dir, mode="all")
 
         print("Writing transcription for auto-validated phrases")
-        self.write_transcriptions(audio_dir, query)
+        self.write_transcriptions(audio_dir, mode="auto-validated")
 
-    def write_transcriptions(self, audio_dir, query):
+    def write_transcriptions(self, audio_dir, mode):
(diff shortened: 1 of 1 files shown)


validation/management/commands/writetranscriptions.py
```diff
@@ -64,40 +82,60 @@ def handle(self, *args, **options):
...
-            from_dir = audio_file
-            to_dir = audio_dir / speaker / "wav" / audio_filename
-            shutil.copyfile(from_dir, to_dir)
+            self.make_directories(training_dir, speaker)
+            self.copy_audio_file(training_dir, audio_file, speaker)
 
             # Treat the transcription for Persephone and save it
             persephone_filename = audio_id + ".txt"
-            persephone_path = audio_dir / speaker / "label" / persephone_filename
+            if "auto-validated" in query:
```
eddieantonio: This is really brittle, imo! I think it's okay for now (no changes needed), but the way I'd do it is that `write_transcriptions` would _not_ take in `query`, but instead take in something like the "kind" (maybe there's a better variable name):

```python
def write_transcriptions(self, audio_dir, kind: Literal["all", "auto-validated"]):
    if kind == "all":
        query = ... # the first query
    elif kind == "auto-validated":
        query = ... # the second query
```

Then this line would be:

```suggestion
            if kind == "auto-validated":
```

Less brittle. But I don't think it _really_ matters in this context.

validation/management/commands/writetranscriptions.py
```diff
@@ -64,40 +82,60 @@ def handle(self, *args, **options):
...
+        for _dir in [
+            speaker_dir,
+            speaker_audio_dir,
+            speaker_persephone_dir,
+            speaker_s4a_dir,
+            speaker_auto_val_dir,
+            speaker_persephone_auto_val_dir,
+            speaker_s4a_auto_val_dir,
+        ]:
+            _dir.mkdir(exist_ok=True)
```
eddieantonio, line 139: It's much nicer to have this as its own function!

eddieantonio (approved):
LGTM! I have a nitpicky comment, but feel free to disregard it!

validation/management/commands/writetranscriptions.py
nienna73: No no, I like this better

eddieantonio (approved):
LGTM!

PR #143, opened 2021-03-05
Header

nienna73:
## What's in this PR:
The header shows "login" when the user is logged out, and "logout" when the user is logged in. The login/out button also still had an inline style, so I pulled that out into its own style class.

**Notes:**
This PR can be merged any time after #138


eddieantonio (approved):
LGTM!

PR #144, opened 2021-03-09
Add Instructor Group

nienna73:
## What's in this PR:
* add a new group called "Instructor"
* users can select their group on registration
* default group is "Community"
* some minor style changes

**Notes:**
Should be merged after #143


andrewdotn (approved):

validation/jinja2/validation/list_phrases.html
```diff
@@ -21,70 +21,80 @@
...
-</div>
-  {% endfor %} {# phrase in phrases #}
+                    {% if phrase.status == "linked" %}
+                        <button class="button button--success-solid translation-judgement-accuracy-yes" data-phrase-id={{ phrase.id }}>Yes</button>
+                        <button class="button button--fail translation-judgement-accuracy-no" data-phrase-id={{ phrase.id }}>No</button></p>
+                    {% else %}
+                        <button class="button button--success translation-judgement-accuracy-yes" data-phrase-id={{ phrase.id }}>Yes</button>
+                        <button class="button button--fail translation-judgement-accuracy-no" data-phrase-id={{ phrase.id }}>No</button></p>
+                    {%  endif %}
+                {% endif %} {# check for instructor privileges #}
```
andrewdotn: Wait—I think the matching `{% if ` tag for this `endif` is on line 53, `{% if is_community %}`

nienna73: Oh, hmm. I updated the comments, do they make more sense now?

andrewdotn: Yeah. Honestly, I’d prefer these end-tag comments not be there at all since it’s so easy for them to get out of sync. But if you find them helpful, that’s much more important than my preferences here.

I am surprised and saddened though that PyCharm doesn’t seem to have a shortcut key to jump from the `endif` or `endfor` to the `if` or `for`.

PR #145, opened 2021-03-10
Audio quality assessment

nienna73:
## What's in this PR:
* when logged in, users can mark a recording as "good" or "bad"
* advanced search for quality and speaker only returns those recordings that match the quality and speaker requirements
* moved cards into their own file

**Notes:**
Quality marking was already being done during the extraction process and is currently included in the production DB

This PR should be merged after #144


validation/views.py
```diff
@@ -444,6 +463,22 @@ def record_translation_judgement(request, phrase_id):
     return JsonResponse({"status": "ok"})
 
 
+@require_http_methods(["POST"])
+def record_audio_quality_judgement(request, recording_id):
+    # TODO: check that user is logged in
+    rec = get_object_or_404(Recording, id=recording_id)
+    print(recording_id)
+    judgement = json.loads(request.body)
```
andrewdotn: Looks like a stray `print()`?

If you like the following suggestion, you’ll need to [`from django.contrib.auth.decorators import login_required`](https://docs.djangoproject.com/en/3.1/topics/auth/default/#the-login-required-decorator)

```suggestion
@login_required()
@require_http_methods(["POST"])
def record_audio_quality_judgement(request, recording_id):
    rec = get_object_or_404(Recording, id=recording_id)
    judgement = json.loads(request.body)
```

nienna73: I removed the extra print--good catch!

I can't seem to get `login_required()` to work. It doesn't seem to realize that I am logged in and it ends up removing all permissions. Maybe we could pair on this a little?

EDIT: oof. I got my branches confused and was logging in with a user that did not, in fact, have any edit permissions. This seems to be working!

andrewdotn (approved):
Does the extra commit mean that `login_required` is now working? If so great! If not, yup, I’m totally up for pairing on it.

andrewdotn:
Oh, I now see your edited comment, as opposed to the original version in the email notification. Great!

PR #148, opened 2021-03-10
fix: exclude bad recordings in API

### LINKED ISSUE #146: Omit bad recordings from legacy API
Anything marked as "bad" should not even APPEAR in the legacy API!

eddieantonio:
Fixes #146


nienna73 (approved):
Beautiful!

PR #150, opened 2021-03-11
Tests

nienna73:
## What's in this PR:
* updated Cypress tests for the new UI
* data-cy tags for all the things I want/need to test

**Notes:**
This is the last of the currently open PRs that should be merged. As in, everything else first, then this.


cypress/integration/basic.spec.js
```diff
@@ -0,0 +1,45 @@
+/// <reference types="cypress" />
```
andrewdotn:
```suggestion
```

This is only needed by old versions of TypeScript.

cypress/integration/basic.spec.js
```diff
@@ -0,0 +1,45 @@
...
+        cy.visit(Cypress.env('home'));
+
+        cy.get('[data-cy="segment-card"]:first')
+            .as('card')
+            .should('be.visible')
+
+        cy.get('[data-cy="card-header"]')
+            .should('be.visible')
+            .within(() => {
+                cy.get('@card')
```
andrewdotn: Something’s funky here, because `() => { blah() }` returns `undefined`. `() => blah()` returns the value of `blah()`, and `() => { return blah() }` also returns the value of `blah()`, but with braces and without an explicit `return`, the return value of `blah()` is ignored.

This [`within` example](https://docs.cypress.io/api/commands/within.html#Forms) indicates you probably want here:

```
cy.get('@card').within(() => {
    cy.get('[data-cy="card-header"]').should('be.visible')
    cy.get('[data-cy="transcription"]').should('be.visible')
    …
})
```

cypress/integration/community.spec.js
```diff
@@ -0,0 +1,64 @@
...
+                },
+            });
+        });
+    })
+
+    it("can mark a translation as good or bad", () => {
+        cy.visit(Cypress.env('home'));
+
+        cy.get('[data-cy="segment-card"]:first')
+            .within(() => {
```
andrewdotn, line 19: Yup, this one looks correct.

cypress/integration/community.spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+                cy.get('[data-cy="yes-button"]')
+                    .should('be.visible')
```
andrewdotn: Cypress “action commands” like `click()` and `type()` (but not `get()`, TIL) already [imply a bunch of assertions](https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Actionability) including that the element is visible.

cypress/integration/instructor.spec.js
```diff
@@ -0,0 +1,32 @@
...
+                body: {
+                    username: "instructor",
+                    password: "1234567890",
+                },
+                headers: {
+                    "X-CSRFTOKEN": token,
+                },
+            });
+        });
+    })
```
andrewdotn: This login bit is copied in a bunch of places, perhaps extract to a shared function?

cypress/integration/community.spec.js
```diff
@@ -0,0 +1,64 @@
...
+                    .click()
+                    .should('have.css', 'background-color', 'rgba(102, 142, 63, 0.5)')
```
andrewdotn: These hard-coded colour values are probably going to be a problem at some point.

Maybe just set the values as constants in a shared JS file, with a comment in both the CSS and JS file saying that if you update one you should update the other too?

Some other quick ideas, which may not be any better than the current thing:

  - Run some JavaScript in Cypress against the page under test to pull out variable definitions [[1]](https://stackoverflow.com/questions/36088655/accessing-a-css-custom-property-aka-css-variable-through-javascript), [[2]](https://github.com/cypress-io/cypress/issues/897)

        cy.window().then((win) => {
            const customGreen = win.getComputedStyle(win.document.documentElement).getPropertyValue('--custom-green');
            // returns " #668E3F"
        })

  - Add some non-displayed divs on the page to grab the CSS value from and
    compare against the target:

        <div
          data-cy="empty-div-with-custom-green-bg"
          style="background: var(--custom-purple); display: none;"
        />

andrewdotn (commented):
Thank you for updating the tests! Tests that pass are much better than broken tests or no tests. I’ve left a few nitpicky comments if you want to have slightly cleaner test code, but if this makes the tests pass, I’m ok with merging as-is.

cypress/integration/basic.spec.js
nienna73: I'm still figuring out when and how to use aliases--this suggestion is much better than what I have!

cypress/integration/community.spec.js
nienna73: Testing for the correct colour is proving to be difficult. I agree that hard coding it is not the way to go. I can store the colour in a variable as suggested, but the variable returns a hex code and the test needs it to be in rgb. I'm wondering if testing for the colour change is even worth it here. What do you think?

andrewdotn: Yeah, testing for an exact colour is probably not worth it. Maybe just check for whether the correct css class is applied? Or if the test is really to make sure that something is reddish or greenish, [this stackoverflow answer](https://stackoverflow.com/a/38963890/14558) actually looks like it *might* be a useful way to map an arbitrary hex / RGB colour into one of eight major colours?

nienna73: I must have missed the `have.class` option when I was first writing this, 'cause that's exactly what I need. No more rgb checking!

cypress/support/commands.js
```diff
@@ -10,7 +10,28 @@
...
+                form: true,
+                body: {
+                    username: username,
+                    password: password,
+                },
+                headers: {
+                    "X-CSRFTOKEN": token,
+                },
+            });
+        });
```
andrewdotn, line 33: This looks good, I am going to borrow it for a PR of mine in itwêwina tomorrow :)

andrewdotn (approved):
Looks good!

cypress/support/commands.js
nienna73: Awesome!

andrewdotn: I was accidentally omitting the password in the POST and the login was failing, so I ended up making sure that the login had actually worked:

```
          followRedirect: false
        }).then(response => {
          expect(response.status).to.eql(302)
          expect(response.headers).to.have.property('location')
          expect(response.headers.location).to.not.contain('login')
        })
```

PR #157, opened 2021-03-19
feat: add failed sessions to ./failed-sessions/!

eddieantonio:
Tries to run all sessions, but moves a symlink to failed sessions if... they failed!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,6 @@ db.sqlite3
 node_modules
 .DS_Store
 crk.zhfst
+
+# Directory of broken dreams:
+failed-sessions/
--- a/librecval/extract_phrases.py
+++ b/librecval/extract_phrases.py
@@ -158,7 +158,9 @@ class RecordingExtractor:
                 failed_dir = Path.cwd() / "failed-sessions"
                 failed_dir.mkdir(exist_ok=True)
                 name = failed_dir / session_id.as_filename()
-                name.symlink_to(session_dir)
+                if not name.exists():
+                    self.logger.error("failed ONCE again: %s", session_dir)
+                    name.symlink_to(session_dir)
                 continue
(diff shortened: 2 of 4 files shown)


tests/test_extract_phrases.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python3
+
```
andrewdotn:
```suggestion
```

librecval/extract_phrases.py
```diff
@@ -140,27 +137,45 @@ def scan(self, root_directory: Path):
...
-            except MissingMetadataError:
-                self.logger.exception("Skipping %s: Missing metadata", session_dir)
+                yield from self.extract_all_recordings_from_session(session_dir)
+            except Exception:
+                session_id = get_session_name_or_none(session_dir)
+                if session_id is None:
+                    continue
+
+                self.logger.exception("Error extracting %s", session_dir)
+                failed_dir = Path.cwd() / "failed-sessions"
```
andrewdotn: This will sometimes end up in random directories if you run the script from random directories. Do you want to build a path off of `Path(__file__).parent`, or maybe some `settings.BASE_DIR` constant?

andrewdotn (approved):

.gitignore
```diff
@@ -115,3 +115,6 @@ db.sqlite3
 node_modules
 .DS_Store
 crk.zhfst
+
+# Directory of broken dreams:
```
dwhieb, line 119: 🎻

librecval/extract_phrases.py
eddieantonio: This package is supposed to be unaware that Django exists, so I'll base it on `Path(__file__).parent`!

PR #158, opened 2021-03-20
Overhaul/organization

### LINKED ISSUE #149: Remove styles from views.py
The styles for the filter buttons are passed in through `views.py` and they shouldn't be.

nienna73:
## What's in this PR:
* filter by session
* display session date next to recording
* removes button style choices that were being made in views.py: Resolves #149 

**Notes:**
I left the commented out "label" style in since I don't know how removing it has or will impact the styles in other branches. I'm waiting to see if any of the labels look wrong in the other branches before removing it.

Some of the styles seem to have gotten confused in the merge--I'll fix that when I make the PR for overhaul/flag.

### DIFF
--- a/validation/jinja2/validation/list_phrases.html
+++ b/validation/jinja2/validation/list_phrases.html
@@ -35,17 +35,18 @@
     </form>
 </div>
 
-    <div>
-        <form action="/action_page.php">
-          <label for="cars">Choose a car:</label>
-          <select name="cars" id="cars">
-            <option value="volvo">Volvo</option>
-            <option value="saab">Saab</option>
-            <option value="opel">Opel</option>
-            <option value="audi">Audi</option>
+    <div class="select">
+        <form method="GET" action="{{
+  url('validation:index')
+}}">
+          <label for="session">Choose a recording session:</label>
+          <select name="session" id="sessions">
+          <option value="all">All Sessions</option>
+              {% for session in sessions %}
+                  <option value="{{ session.date }}">{{ session.date }}</option>
+              {% endfor %}
           </select>
-          <br><br>
-          <input type="submit" value="Submit">
+          <input type="submit" value="Select" class="button button--success button--small">
         </form>
     </div>
 
--- a/validation/jinja2/validation/list_phrases.html
+++ b/validation/jinja2/validation/list_phrases.html
@@ -28,41 +28,45 @@
             session_mode.value = params.get('session');
         }
     </script>
-<div>
-  <form class="form-inline ml-2 my-2 my-lg-0" method="GET" action="{{
+<div class="center">
+  <form method="GET" action="{{
     url('validation:index')
   }}">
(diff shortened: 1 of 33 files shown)


validation/jinja2/validation/list_phrases.html
```diff
@@ -21,26 +21,47 @@
...
+  <form method="GET" action="{{
     url('validation:index')
   }}">
     <div class="filter">
-      <button class="{{ all_class }}" type="submit" id="filter-all" name="mode" value="all">All</button>
-      <button class="{{ new_class }}" type="submit" id="filter-new" name="mode" value="new">New</button>
-      <button class="{{ linked_class }}" type="submit" id="filter-linked" name="mode" value="linked">Linked</button>
+        Filter by:
+        <input type="radio" id="filter-all" name="mode" value="all">
+        <label for="all"> All</label>
```
andrewdotn: I think the `for` is supposed to match `id`, i.e., `for="filter-all"`. Does clicking the label select the radio button, or do you have to click the little circle of the radio button itself?

validation/views.py
```diff
@@ -495,6 +502,20 @@ def user_is_linguist(user):
     return False
 
 
+def get_phrases_from_session(session, all_phrases):
+    phrases_from_session = []
+    session_date = datetime.datetime.strptime(session, "%Y-%m-%d").date()
+    for phrase in all_phrases:
+        for recording in phrase.recordings:
+            if (recording.session.date == session_date) and (
+                phrase not in phrases_from_session
```
andrewdotn: If there might be thousands of phrases, this could get slow, because checking if `phrase` is in the growing `phrases_from_session` list is O(n^2) whereas it could be only O(n) if you used a set.

validation/jinja2/validation/list_phrases.html
nienna73: You are correct! Clicking the label didn't work before, but after changing this, you can click the label or directly on the little circle.

validation/views.py
nienna73: I was worried about this being very slow on production so I'll switch it over to a set.

andrewdotn: Now that I take a second look, there *should* be a way to get this out of the database directly with some sort of query…

nienna73: I was really hoping there would be, but I couldn't find a direct link between all three tables. And I don't think I can index across multiple tables

andrewdotn: It doesn’t match the name of the method if there can be multiple sessions per date, but finding all phrases from all sessions on a given date should be `Phrase.objects.filter(recording__session__date='2015-03-23')`

nienna73: This is great--so much nicer to look at! Thanks for figuring that out.

andrewdotn (approved):

validation/views.py
andrewdotn: :)

PR #160, opened 2021-03-23
Flag

nienna73:
## What's in this PR:
* users can now flag an issue!
* must be logged in to flag an issue
* made a new Issue model to save issues that have been reported
* some small adjustments to button styles
* ALSO added "I don't know" button
* ALSO added django dev tools to this repo

This is what it looks like:
Screen Shot 2021-03-23 at 4 00 15 PM

In addition to the information shown in the popup, an Issue stores a reference to the Phrase with which its associated, the username of the user that filed the issue, and the date on which the issue was filed.

**Notes:**
The "comments" section on the popup is not mandatory. The idea with this section is that users will leave a few notes as to why they're flagging the issue, but I don't want this to become a hindrance to users working their way through many entries in one sitting.

@dwhieb I would appreciate your input on whether or not to make the comment field mandatory as well as the overall flow of this popup

@aarppe I would appreciate your input on the wording used in the popup itself: should I change any of the prompts? Did I miss any reasons why someone might flag an entry?


andrewdotn (approved):

aarppe:
If one comments that the English translation or Cree wordform are not the best, then we should ask them to provide such a better translation or wordform, at that time. That is, I don't think it makes the most sense to provide a negative assessment of either, without providing a better option then and there. For the English translation, this would be a matter of fixing that in the appropriate pop-up. For the better Cree word, that would amount to a suggestion to record that better word.

But I realize this would best be considered in the context of the entire workflow, and my views could change in such a consideration.

nienna73:
I think I'll make the "comment" mandatory for now and add some helptext about what we expect people to write in there.

validation/models.py
```diff
@@ -403,6 +404,48 @@ def get_path_to_audio_directory() -> Path:
...
+    phrase = models.ForeignKey(Phrase, on_delete=models.CASCADE)
+
+    created_by = models.CharField(
+        help_text="The person who filed this issue",
+        default="",
+        max_length=64,
+    )
+
+    created_on = models.DateField(
+        help_text="When was this issue filed?", default=timezone.now()
```
eddieantonio: This line sets `created_on` to the date ~~_when the server was started_~~ **when the migration was run**. Luckily, Django provides `auto_now_add`, which I believe is the intention of the line:

```suggestion
        help_text="When was this issue filed?",
        auto_now_add=True,
```

See: https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.DateField.auto_now_add

validation/static/index.js
```diff
@@ -47,21 +47,46 @@ document.addEventListener('DOMContentLoaded', () => {
...
-                                h.classList.add('card__header--grey')
-                            }
+                    for (let h of headers) {
+                        if (h.dataset.phraseId === phraseId) {
+                            h.classList.remove('card__header--green')
+                            h.classList.add('card__header--grey')
                         }
+                    }
                     
+                } else if (judgement === "idk") {
```
eddieantonio, line 65: Love this. 😂

validation/migrations/0004_auto_20210326_1039.py
```diff
@@ -0,0 +1,37 @@
...
+            name="created_by",
+            field=models.CharField(
+                default="", help_text="The person who filed this issue", max_length=64
+            ),
+        ),
+        migrations.AddField(
+            model_name="issue",
+            name="created_on",
+            field=models.DateField(
+                default=datetime.datetime(2021, 3, 26, 16, 39, 46, 988947, tzinfo=utc),
```
eddieantonio: This is the consequence of setting `Issue.created_on`  to `timezone.now()`. 🙈

validation/jinja2/validation/_segment_card.html
```diff
@@ -56,41 +61,65 @@
...
+                        </div>
+                        <div class="modal-footer">
+                            <input data-cy="cancel-button" type="button" class="button button--fail" data-dismiss="modal" value="Cancel">
+                            <input data-cy="save-button" type="submit" class="button button--success" value="Save">
+                        </div>
+                    </form>
+                </div>
             </div>
-        </article>
+        </div>
```
eddieantonio: Consider using `<dialog>` instead: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

There is a way to mix this with `<form method="dialog">` to get some modal functionality "for free", but I don't know how to coax Django's generated `Form` classes with it.

validation/jinja2/validation/_segment_card.html
```diff
@@ -56,41 +61,65 @@
...
+                    </div>
+
+                    <form action="" method="post">
+                        <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}">
+                        <div class="modal-body">
+                            {{ forms[phrase.id] }}
+                        </div>
+                        <div class="modal-footer">
+                            <input data-cy="cancel-button" type="button" class="button button--fail" data-dismiss="modal" value="Cancel">
+                            <input data-cy="save-button" type="submit" class="button button--success" value="Save">
```
eddieantonio: These are probably better as `<button type="button" form=...>` but I don't know how to get the ID from the `form[phrase-id]`.

validation/views.py
```diff
@@ -512,3 +532,43 @@ def user_is_expert(user):
...
+    new_issue = Issue(
+        phrase=phrase,
+        other="other" in issues,
+        bad_cree="bad_cree" in issues,
+        bad_english="bad_english" in issues,
+        bad_recording="bad_rec" in issues,
+        comment=comment,
+        other_reason=other_reason,
+        created_by=user,
+        created_on=datetime.datetime.now(),
```
eddieantonio, line 578: You can just delete the default in the `Issue` model if you use this every time.

eddieantonio (commented):
If `created_on` should be set to the data that the `Issue` is created on, and not the date in the migration, then changes are needed.

validation/models.py
nienna73: Ah, I think I added this because I already had some Issues locally, but didn't have the date on them. I can change it to auto_now_add, or I think you mention later in the comments that I can just take it out. Either way, this doesn't need to be in production--good catch!

validation/jinja2/validation/_segment_card.html
nienna73: `<dialog>` is not supported on Firefox, IE, or Safari, so I don't actually think that's what I want to use.

validation/jinja2/validation/_segment_card.html
nienna73: I won't! There's an actual form element wrapper around the django form, so I'll give that an ID that's dependent on the phrase ID and access it that way in the buttons.

nienna73:
I think I caught all the comments and I updated the Cypress tests, too!

cypress/integration/expert.spec.js
```diff
@@ -2,7 +2,8 @@
 
 describe("Community members", () => {
     beforeEach(() => {
-        cy.login("community", "1234567890");
+        cy.login("expert", "1234567890");
+        cy.get('#djHideToolBarButton').click();
```
eddieantonio, line 6: In a future PR, it might be useful to create a new [Cypress custom command](https://docs.cypress.io/api/cypress-api/custom-commands) to hide the debug toolbar.

validation/models.py
```diff
@@ -25,6 +25,7 @@
 from django.db import models
 from django.db.models.signals import pre_save
 from django.dispatch import receiver
+from django.utils import timezone
```
eddieantonio: Unused import now:

```suggestion
```

eddieantonio (approved):
LGTM!

You may want to refactor things in a future PR (e.g., extract function, extract templates), just so that extending the existing codebase becomes easier!

PR #161, opened 2021-03-23
Update README

nienna73:
## What's in this PR:

- Removed all references to TextGrid
- Added instructions for Cypress tests
- Updated instructions for generating transcription files
- Added my email to the License
- Other small spelling fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,12 +320,8 @@ npx cypress open
 This opens the interactive Cypress testing window. Select `Run integration specs` to run 
 all tests.
 
-> **Note:** These tests will run against whichever database is present. Meaning if you're developing
-> locally, they will use that database; if you're running on production, they will use that database. **DO NOT** run
-> Cypress tests on Production: this will validate some entries due to the nature of the tests.
-> 
-> Additionally, at least 1 test will always fail: either one of the `register` tests or one of the `login` 
-> tests. This is because a) the username is already taken or b) the user does not exist. This is fine.
+**Note: DO NOT** try to run Cypress tests on production: this will validate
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -302,6 +310,22 @@ This will run the [mypy][] static type checker, and then [pytest][].
...
+```
+This opens the interactive Cypress testing window. Select `Run integration specs` to run 
+all tests.
+
+> **Note:** These tests will run against whichever database is present. Meaning if you're developing
+> locally, they will use that database; if you're running on production, they will use that database. **DO NOT** run
+> Cypress tests on Production: this will validate some entries due to the nature of the tests.
+> 
+> Additionally, at least 1 test will always fail: either one of the `register` tests or one of the `login` 
+> tests. This is because a) the username is already taken or b) the user does not exist. This is fine.
```
andrewdotn: Is there a way to:
 1. Have cypress verify that it is *not* running in production
 2. Use a randomly-generated username so that the tests pass normally?

nienna73: Ideally, I would have Cypress tests running on a test database so I could guarantee which users exist and what permissions they have, but I don't know how to do that and it's not near the top of my list of priorities.

andrewdotn: Ok, I understand. My main concern was that the wording suggested the possibility of running against production before saying not to; how about shortening it to the warning:

```suggestion
> **Note: DO NOT** try to run Cypress tests on production: this will validate
> some entries due to the nature of the tests.
```

nienna73: Perfect, I'll change it so it's clear.

PR #164, opened 2021-03-25
Merge into Production

nienna73:
## What's in this PR:
* the entire UI overhaul
* better error catching in the extraction process
* unit tests are up to date
* Cypress tests are up to date
* auto-validation
* better CSS class names

**Notes:**
All of these changes have been individually reviewed as they were merged into development, so this final merge mostly needs a sanity check to make sure nothing fell through the cracks.


Pipfile
```diff
@@ -18,6 +18,8 @@ pympi-ling = "*"
 requests = "*"
 uwsgi = "*"
 divvunspell-python = {ref = "3aa656d36a2fa1ede90e53c320b2095c05cf7c05", git = "https://github.com/divvun/divvunspell-sdk-python"}
+black = "==20.8b1"
```
andrewdotn: This might be happier in `[dev-packages]`.

cypress/integration/expert.spec.js
```diff
@@ -0,0 +1,47 @@
...
+    })
+
+    it("can mark a translation as good or bad", () => {
+        cy.visit(Cypress.env('home'));
+
+        let customGreen = null;
+
+        cy.window()
+            .then((win) => {
+                customGreen = win.getComputedStyle(win.document.documentElement).getPropertyValue('--custom-pale-green');
```
andrewdotn: This is never used. I’m glad we didn’t have to go this route.

cypress/integration/expert.spec.js
```diff
@@ -0,0 +1,47 @@
...
+
+    it("can mark a recording as good or bad", () => {
+        cy.visit(Cypress.env('home'));
+
+        cy.get('[data-cy="segment-card"]:first')
+            .within(() => {
+
+                cy.get('[data-cy="good-button"]:first')
+                    .click()
+                    .should('have.css', 'background-color', 'rgba(102, 142, 63, 0.5)')
```
andrewdotn: I think you were happier using `have.class` instead for this sort of thing?

cypress/support/commands.js
```diff
@@ -10,7 +10,28 @@
...
+                url: Cypress.env("login_url"),
+                form: true,
+                body: {
+                    username: username,
+                    password: password,
+                },
+                headers: {
+                    "X-CSRFTOKEN": token,
+                },
+            });
```
andrewdotn: I ran into some trouble with [some very similar code](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/master/cypress/integration/admin.spec.js#L29-L33): if the POST fails for some reasons, like an invalid password, then the `login()` method still succeeds, and debugging the failing parts of the test after `login()` gets very confusing, because the user is not logged in.

validation/management/commands/autoval.py
```diff
@@ -0,0 +1,53 @@
...
+    help = "auto-validates phrases"
+
+    def handle(self, *args, **options):
+        """
+        Iterates over all the phrases in the database
+        and auto-standardizes wordforms where there is exactly one
+        matching analysis
+        """
+        phrases = Phrase.objects.all()
+        for phrase in tqdm(phrases):
```
andrewdotn: If there are a *lot* of phrases and takes a long time to even get started, `phrases.iterator()` goes through phrases in batches of 1000 instead of loading them all in memory up front.

validation/management/commands/writetranscriptions.py
```diff
@@ -64,55 +83,76 @@ def handle(self, *args, **options):
...
+
+    def make_directories(self, training_dir, speaker):
+        # Create necessary directories if they do not exist
+        speaker_dir = training_dir / speaker
+        speaker_audio_dir = training_dir / speaker / "wav"
+        speaker_persephone_dir = training_dir / speaker / "label"
+        speaker_s4a_dir = training_dir / speaker / "s4a"
+        speaker_auto_val_dir = training_dir / speaker / "auto_val"
+        speaker_persephone_auto_val_dir = training_dir / speaker / "auto_val" / "label"
+        speaker_s4a_auto_val_dir = training_dir / speaker / "auto_val" / "s4a"
```
andrewdotn:
```suggestion
        speaker_audio_dir = speaker_dir / "wav"
        speaker_persephone_dir = speaker_dir / "label"
        speaker_s4a_dir = speaker_dir / "s4a"
        speaker_auto_val_dir = speaker_dir / "auto_val"
        speaker_persephone_auto_val_dir = speaker_auto_val_dir / "label"
        speaker_s4a_auto_val_dir = speaker_auto_val_dir / "s4a"
```

validation/static/index.js
```diff
@@ -0,0 +1,143 @@
+"use strict";
+
+//    translation-judgement-accuracy-yes
```
andrewdotn:
```suggestion
```

validation/static/index.js
```diff
@@ -0,0 +1,143 @@
...
+                    headers: {
+                        'Content-Type': 'application/json',
+                        'X-CSRFToken': csrftoken
+                    },
+                    body: JSON.stringify({judgement})
+                })
+
+                let r = await response.json()
+
+                if (r.status != 'ok') {
```
andrewdotn, line 21: I recently learned that `fetch` responses have a simpler built-in property to check if ok or not:

```suggestion
                if (!r.ok) {
```

andrewdotn (approved):
This looks good overall. I’ve made a few suggestions but they are all optional.

validation/views.py
```diff
@@ -372,6 +485,102 @@ def register(request):
...
 def encode_query_with_page(query, page):
     query["page"] = page
     return f"?{query.urlencode()}"
+
+
+def user_is_linguist(user):
+    if user.is_authenticated:
+        for g in user.groups.all():
+            if g.name == "Linguist":
+                return True
```
andrewdotn: This whole method can be replaced with `user.groups.filter(name='Linguist').exists()`, and the next one with `user.groups.filter(name__in=['Linguist', 'Expert']).exists()`. If the user is not authenticated, `user` is an [`AnonymousUser`](https://docs.djangoproject.com/en/3.1/ref/contrib/auth/#anonymoususer-object) object on which `user.groups` will always return an empty queryset, and `False` for `.exists()`.

cypress/integration/expert.spec.js
nienna73: Oh my goodness, me too! Taking this out now

cypress/support/commands.js
nienna73: We fixed this on one of the CI branches so it'll show up here soon!

validation/management/commands/autoval.py
nienna73: There are _a lot_ of phrases! I'll add this is now, but sadly we won't see how much faster it makes it for a long time...if ever.

nienna73: Hmm this seems to disrupt `tqdm` somehow, though. It doesn't show the full progress bar when I add the `.iterator()`

validation/static/index.js
nienna73: This doesn't seem to work in this context 🤔

validation/views.py
nienna73: Nice!

validation/management/commands/autoval.py
andrewdotn: Oops, sorry, [the way I do this is](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/21018addd9e9b5921c27f26b9d7eb4a5a0b2105f/CreeDictionary/phrase_translate/management/commands/translatewordforms.py#L150-L151),

    for blah in tqdm(queryset.iterator(), total=queryset.count()):

This does the `count()` once at the start, which is faster than retrieving everything and jamming it all into memory, then page through in batches of 1000. If the database was actively being written to, the initial count could be different from how many items you actually iterate through, but if all that affects is the progress meter, no biggie.

validation/static/index.js
andrewdotn: That’s sad. I’ve seen it [in the docs](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) and want to try it but haven’t yet used it myself. 🤷

validation/management/commands/autoval.py
nienna73: Perfect, that does it. I can't tell the difference locally since I have <1000 entries on my laptop, but I'm sure it'll make a difference when running this against the entire DB.

ci_run_cypress.py
```diff
@@ -0,0 +1,36 @@
+from subprocess import check_call, Popen
```
eddieantonio, line 9: I would appreciate a docstring explaning this file and its behaviour; perhaps incorporate this into argparse's `--help` text. e.g.,

 - creates a new test db, deleting any existing test db if it exists
 - can pass `--interactive` to open Cypress's interactive mode rather than running all the tests in headless mode

cypress.json
```diff
@@ -5,6 +5,6 @@
         "login_url": "/login",
         "register_url": "/register",
         "logout_url": "/logout",
-        "segment_details_url": "/segment/1"
+        "segment_details_url": "/segment/6"
```
eddieantonio, line 8: UH, WHY?

validation/management/commands/ensuretestdb.py
```diff
@@ -0,0 +1,50 @@
...
+    will be migrated. If assorted other things need to be in there, they will be
+    added if missing.
+    """
+
+    def handle(self, *args, **options):
+
+        call_command("migrate", verbosity=0)
+
+        import_test_phrases()
+        # stub_recordings()
```
eddieantonio: Why is this commented out?

validation/management/commands/ensuretestdb.py
```diff
@@ -0,0 +1,50 @@
...
+    user, _ = User.objects.get_or_create(username=username)
+    user.set_password("1234567890")
+    user.save()
+    group, _ = Group.objects.get_or_create(name=group_name)
+    group.user_set.add(user)
+
+
+def create_test_users():
+    for username in ["expert", "instructor", "linguist", "learner"]:
+        create_test_user(username, username.title())
```
eddieantonio, line 49: This is nice!

validation/management/commands/autoval.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        phrases = Phrase.objects.all()
+        for phrase in tqdm(phrases.iterator(), total=phrases.count()):
+            if not phrase.validated:
+                segment_name = phrase.field_transcription
+                suggestions = get_distance_with_translations(segment_name)
+                match = perfect_match(segment_name, suggestions)
+
+                # only save the analysis if there is exactly one
+                if exactly_one_analysis(match):
```
eddieantonio, line 48: These function names are still 🔥 🔥 🔥

validation/static/index.js
eddieantonio: Because`r` is the parsed HTTP body (JSON returned from the API); `response` is the HTTP response!

But that means `status` is part of our speech-db API! So I'm not sure if this is appropriate:

```suggestion
                if (response.status != 'ok') {
```

eddieantonio (approved):
LGTM! I would just like docs in the integration test runner script.

cypress.json
eddieantonio: Pretty sure this is “awas”!

ci_run_cypress.py
```diff
@@ -17,11 +27,17 @@ def modified_env(**kwargs):
 TEST_DB = "test_db.sqlite"
 TEST_DB_FILE = Path(TEST_DB)
 
+# We delete the DB before we run the tests to make sure there
+# aren't any extra users in the DB (i.e. the ones made when
+# testing the register process)
```
eddieantonio, line 32: This is valuable to know!

eddieantonio (approved):
LGTM! (again!)

validation/management/commands/ensuretestdb.py
nienna73: Oops. I thought I was going to need another function to do this, turns out I just needed a fixture. It's gone now!

PR #167, opened 2021-03-30
Cypress tests on CI

nienna73:
## What's in this PR:
* Cypress tests run on CI
* create a test DB for Cypress tests
* added fixtures to build the test DB

**Notes:**
I removed the checks for specific CSS values in the Cypress tests. These tests were temperamental and failed in GitHub Actions. Also, checking CSS isn't really the purpose of Cypress.

### DIFF
--- a/cypress/integration/edit_segment.spec.js
+++ b/cypress/integration/edit_segment.spec.js
@@ -115,50 +115,6 @@ describe("Edit segment", () => {
             })
     })
 
-    // The following test only works if the entries have a revision history
-    // Since the test-db is made when the tests are run, there is no revision history
-    // to test against
-    it.skip("should load content when clicking Revert", () => {
-        cy.visit(Cypress.env("segment_details_url"));
-
-        cy.get("#revision-table").within(() => {
-            cy.get('[data-cy="revision-transcription"]')
-                .first()
-                .invoke('text')
-                .as('transcription')
-            cy.get('[data-cy="revision-translation"]')
-                .first()
-                .invoke('text')
-                .as('translation')
-            cy.get('[data-cy="revision-analysis"]')
-                .first()
-                .invoke('text')
-                .as('analysis')
-            cy.get('input:first')
-                .should('have.value', 'Revert')
-                .click()
-        })
-
-        cy.get('[data-cy=edit-div]').should('be.visible')
-        cy.get('#id_cree')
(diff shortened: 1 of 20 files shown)


recvalsite/settings.py
```diff
@@ -33,6 +33,7 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = config("DEBUG", default=False, cast=bool)
+DEBUG_DJANGO = ""
```
andrewdotn:
```suggestion
DEBUG_DJANGO = config("DEBUG_DJANGO", default=DEBUG, cast=bool)
```

?

You might also want to rename this to `USE_DJANGO_DEBUG_TOOLBAR`

cypress/integration/edit_segment.spec.js
```diff
@@ -122,7 +115,10 @@ describe("Edit segment", () => {
             })
     })
 
-    it("should load content when clicking Revert", () => {
+    // The following test only works if the entries have a revision history
+    // Since the test-db is made when the tests are run, there is no revision history
+    // to test against
+    it.skip("should load content when clicking Revert", () => {
```
andrewdotn: Could you test the revert part after having cypress make a test edit?

andrewdotn (approved):
I have a few suggestions, but since the tests pass in GitHub Actions, this is great, and I don’t want to hold that up.

cypress/integration/edit_segment.spec.js
nienna73: Likely, yes. Let me try

nienna73: Great suggestion! Putting the revert test after the "save on edit" test solved the problem.

recvalsite/settings.py
```diff
@@ -33,7 +33,7 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = config("DEBUG", default=False, cast=bool)
-DEBUG_DJANGO = ""
+USE_DJANGO_DEBUG_TOOLBAR = config("DEBUG_DJANGO", default=DEBUG, cast=bool)
```
andrewdotn, line 36: You’ll want `USE_DJANGO_DEBUG_TOOLBAR` on the right side too, and in the `ci_run_cypress` script.

PR #168, opened 2021-03-31
fix: regenerated Pipfile.lock

nienna73:
Deploy failed 'cause Pipfile.lock was out of date. I regenerated it so hopefully it'll work now.


eddieantonio (approved):
I was able to succesfully build the Docker image from this — so I think it should work!

PR #170, opened 2021-04-01
Some emergency fixes to get the site working

andrewdotn:
- Use DB from local volume, not a network share

  - Create folder for static content, and with required permissions

  - Use ten processes instead of whatever the uwsgi default is

  - Copy-paste some logging config from itwêwina and hope it helps?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,8 @@ ARG WSGI_USER=speech-db
 RUN groupadd --system --gid ${UID_GID} ${WSGI_USER} \
  && useradd --no-log-init --system --gid ${WSGI_USER} --uid ${UID_GID} ${WSGI_USER} \
  && mkdir /app \
- && chown ${WSGI_USER}:${WSGI_USER} /app
+ && mkdir -p /var/www/recvalsite \
+ && chown ${WSGI_USER}:${WSGI_USER} /app /var/www/recvalsite
 
 USER ${WSGI_USER}:${WSGI_USER}
 WORKDIR /app/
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -42,8 +43,8 @@ COPY --chown=${WSGI_USER}:${WSGI_USER} . .
 
 ENV VIRTUAL_ENV="/app/.venv"
 ENV PATH="${VIRTUAL_ENV}/bin:${PATH}"
-RUN mkdir static && python manage.py collectstatic
+RUN python manage.py collectstatic
 
 EXPOSE 8000
 ENV UWSGI_HTTP=:8000 UWSGI_MASTER=1 UWSGI_HTTP_KEEPALIVE=1 UWSGI_AUTO_CHUNKED=1 UWSGI_WSGI_ENV_BEHAVIOUR=holy
-CMD ["uwsgi", "-w", "recvalsite.wsgi", "--static-map", "/static=/var/www/recvalsite/static"]
+CMD ["uwsgi", "-w", "recvalsite.wsgi", "--processes", "10", "--static-map", "/static=/var/www/recvalsite/static"]
```
nienna73, line 50: I think there's also a place in the nginx file where we could make this change? It's probably fine here for now though

docker-compose.yml
```diff
@@ -8,5 +8,5 @@ services:
         volumes:
         # HOST : CONTAINER
         - "./.env:/app/.env"
-        - "/data/application-data/speech-db/db.sqlite3:/app/db.sqlite3"
-        - "/data/application-data/speech-db/audio:/app/data/audio"
\ No newline at end of file
+        - "./db.sqlite3:/app/db.sqlite3"
```
nienna73, line 11: If this isn't a mounted volume anymore, it likely doesn't need to be in here, right?

nienna73 (approved):
A couple minor comments, but neither impact functionality. Thanks for fixing this!

docker-compose.yml
andrewdotn: We definitely need this; otherwise the container won’t have a database to work with on startup, and if you do manage to get some data saved into the container, it’ll be lost when a new container is created on the next deploy.

Mounting a file or directory into a container is similar but different from mounting a network share. `/data` on itw/kor is a network share mounted from the gw machine. When only a docker mount is involved, the process is much more reliable than going over a network, but you still run into some issues that you wouldn’t see if no extra levels of mounting were involved, e.g., some of the stuff we see with UIDs and permissions.

nienna73: Oh of course, I didn't see the `:` when I made the comment

Dockerfile
andrewdotn: itwewina is set up to use a [`uwsgi.ini`](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/master/CreeDictionary/uwsgi.ini) file for [almost all](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/master/docker/Dockerfile#L107) the uwsgi settings; this is the “let me try to get the site working as quickly as possible” version.

I’ll make a note to clean this up later.

PR #171, opened 2021-04-01
Changes to queries for efficiency

nienna73:
## What's in this PR:
* better use of Django's query abilities to try and speed up query times

**Notes:**
I want to push this one straight to production in the hopes that it'll speed up load times.

### DIFF
--- a/validation/views.py
+++ b/validation/views.py
@@ -20,6 +20,7 @@ import datetime
 import io
 import json
 import operator
+from functools import reduce
 from pathlib import Path
 
 from django.conf import settings
@@ -41,6 +42,7 @@ from django.contrib.auth.models import User, Group
 from django.contrib.auth import authenticate, login as django_login
 from django.views.decorators.http import require_http_methods
 from django.contrib.auth.decorators import login_required
+from django.db.models import Q
 
 from librecval.normalization import to_indexable_form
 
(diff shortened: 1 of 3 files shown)


validation/views.py
```diff
@@ -210,58 +200,42 @@ def advanced_search_results(request):
...
+    if speakers and "all" not in speakers:
+        recordings_filter_query.append(~Q(speaker__in=speakers))
+    if quality and "all" not in quality:
+        recordings_filter_query.append(~Q(quality__in=quality))
+
+    for phrase in phrase_matches:
+        if recordings_filter_query:
+            recordings[phrase] = phrase.recordings.exclude(
+                reduce(operator.and_, recordings_filter_query)
+            )
```
andrewdotn: Since this is an `exclude` query, not a `filter` one, I would like to suggest that the name should match:

```suggestion
    recordings_exclude_query = []
    if speakers and "all" not in speakers:
        recordings_exclude_query.append(~Q(speaker__in=speakers))
    if quality and "all" not in quality:
        recordings_exclude_query.append(~Q(quality__in=quality))

    for phrase in phrase_matches:
        if recordings_exclude_query:
            recordings[phrase] = phrase.recordings.exclude(
                reduce(operator.and_, recordings_exclude_query)
            )
```

validation/views.py
```diff
@@ -67,43 +69,34 @@ def index(request):
             all_phrases = Phrase.objects.exclude(status="auto-validated").order_by(
                 "transcription"
             )
```
andrewdotn: This suggestion is explained in the next comment.

```suggestion
            all_phrases = Phrase.objects.exclude(status="auto-validated")
```

validation/views.py
```diff
@@ -67,43 +69,34 @@ def index(request):
...
-        all_phrases = Phrase.objects.filter(status="linked").order_by("transcription")
-    elif mode == "auto-validated":
-        all_phrases = Phrase.objects.filter(status="auto-validated").order_by(
-            "transcription"
-        )
+    else:
+        all_phrases = Phrase.objects.filter(status=mode).order_by("transcription")
 
     all_phrases = all_phrases.prefetch_related("recording_set__speaker")
+    all_phrases = sorted(all_phrases, key=operator.attrgetter("transcription"))
```
andrewdotn: How about applying `order_by` once at the end, and letting the sort happen only on the database end?

```suggestion
    else:
        all_phrases = Phrase.objects.filter(status=mode)

    all_phrases = all_phrases.order_by('transcription').prefetch_related("recording_set__speaker")
```

validation/views.py
nienna73: Sure thing!

andrewdotn (approved):
Nice, I like how you used `reduce` to `|` everything together.

If the query starts to get hard to manage, it’s sometimes helpful to add unit tests that create some tests objects and validates that the query method returns exactly the right objects for the various different input options. But I think that gets into additional complexity around creating a temporary empty database for a single unit test.

validation/views.py
nienna73: I tried to take out the `sorted` and I couldn't figure out why that changed the results. The two competing `order_by` calls would do it though--thanks!

andrewdotn (approved):

PR #179, opened 2021-04-08
More options

nienna73:
## What's in this PR:
* added "Wrong Speaker" button
* pressing "Wrong Speaker" creates a new Issue linked to the recording saying that it has the wrong speaker
* removed "Specify reason" from modal
* removed "other_reason" from Issue model
* added "Cree suggestion" to the modal and the Issue model
* added "English suggestion" to the modal and the Issue model
* turned "Comments" field into a text area to there's more space

**Notes:**
We'll have to be careful about pushing this one to production as it requires some database migrations. I'll double check the contents of the validation_issue table before doing anything to make sure we don't lose info from already-filed issues (if there are any).


andrewdotn (approved):
Looks good

If you’re worried about the db, make a copy of it before merging

(The deploy script could actually do that automatically, I have set that up with other projects in the past, that doing a deploy automatically backs up the database as the first step.)

nienna73:
That's a good suggestion! I'll add that to my list of to-dos for this project.

PR #182, opened 2021-04-12
mount sessions to docker container

nienna73:
## What's in this PR:
* show Docker where the sessions folder is

I'm trying to run the Persephone transcription-writing process, but first I need to have the audio files in a .wav format. In order to do that, I need to have access to the sessions.

**Notes:**
This PR is going straight to Production


andrewdotn (approved):
Looks good to me!

BTW if you are mounting some data that is specifically supposed to be read-only, you can add `:ro` to the end, i.e.,

     - "/data/maskwacis-recordings/sessions:/app/data/sessions:ro"

though in this case I expect normal UNIX permissions should take care of things.

PR #183, opened 2021-04-13
Make Issue phrase+recording admin fields readonly

andrewdotn:
Requesting the speaker name and so on for every recording to show a nice
`__str__()` for everything in the select box was sending way too many
requests to the database.

There are some more sophisticated ways to address this here:

https://stackoverflow.com/questions/6892906/how-to-force-django-admin-to-use-select-related

### DIFF
--- a/validation/admin.py
+++ b/validation/admin.py
@@ -22,4 +22,8 @@ admin.site.register(Phrase, SimpleHistoryAdmin)
 admin.site.register(RecordingSession)
 admin.site.register(Speaker)  # TODO: use simplehistory
 admin.site.register(Recording, SimpleHistoryAdmin)
-admin.site.register(Issue)
+
+class IssueAdmin(admin.ModelAdmin):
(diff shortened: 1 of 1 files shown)


nienna73 (approved):

eddieantonio:
Wild! So when "phrase" and "recording" were modifiable, Django would make WAY TOO MANY DATABASE QUERIES to the point where nginx would timeout the request? Am I understanding this right?

nienna73:
That's exactly it, yes. It was loading **all** the Phrases, Speakers, and Recordings so it could present them as options to be modified.

andrewdotn:
There’d be a dropdown for Recording, populated with every Recording, and since `Recoding.__str__` is 

        return f'"{self.phrase}" recorded by {self.speaker} during {self.session}'

stringifying each Recording to fill out the dropdown would require loading the associated phrase, speaker, and session objects for every recording, one at a time.

PR #184, opened 2021-04-14
Options button in header

nienna73:
## What's in this PR:
* added an "Options" button to the header
* added the option to show <e> with or without the macron

### DIFF
--- a/validation/jinja2/validation/_base.html
+++ b/validation/jinja2/validation/_base.html
@@ -51,11 +51,12 @@ const csrftoken = document.querySelector('[name=csrfmiddlewaretoken]').value;
     {# Login button #}
     <div>
         <button class="button--none" onclick="showMenu()"><span class="menu__caret">Options <i class="fa fa-caret-down"></i></span></button>
-        {{ menu.menu_options(request, auth, showMacron) }}
+
     </div>
   </nav>
 
   <main role="main" class="container">
+    {{ menu.menu_options(request, auth, showMacron) }}
     {% block content %}
     <p class="error">
       If you're seeing this, something is wrong. Please contact
--- a/validation/jinja2/validation/_base.html
+++ b/validation/jinja2/validation/_base.html
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
   <link rel="icon" type="image/png" href="/static/images/little-m.png">
-  <link rel="stylesheet" href="/static/styles.css" type="text/css">
+  <link rel="stylesheet" href="{{ static('styles.css') }}" type="text/css">
   <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.0/css/bootstrap.min.css" integrity="sha384-9gVQ4dYFwwWSjIDZnLEWnxCjeSWFphJiwGPXr1jddIhOegiu1FwO5qRGvFXOdJZ4" crossorigin="anonymous">
   <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
   <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/js/bootstrap.min.js" integrity="sha384-JZR6Spejh4U02d8jOt6vLEHfe/JQGiRRSQQxSfFWpi1MquVdAyjUar5+76PVCmYl" crossorigin="anonymous"></script>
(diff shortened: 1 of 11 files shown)


validation/jinja2/validation/_base.html
```diff
@@ -14,12 +15,15 @@
 const csrftoken = document.querySelector('[name=csrfmiddlewaretoken]').value;
 </script>
       <script src="/static/index.js"></script>
+      <script src="/static/menu.js"></script>
```
eddieantonio: I don't understand what's going on with the static stuff, but this is the _cromulent_ way of including static files:

```suggestion
      <script src="{% static 'menu.js' }%"></script>
```

Please see: https://docs.djangoproject.com/en/3.2/howto/static-files/#configuring-static-files

validation/static/menu.js
```diff
@@ -0,0 +1,20 @@
...
+        menu.classList.remove("menu__none");
+        menu.classList.add("menu__block");
+    } else {
+        menu.classList.add("menu__none");
+        menu.classList.remove("menu__block");
+    }
+}
+
+function showMacron(option) {
+    document.cookie = "macron=" + option + "; SameSite=Lax;"
```
eddieantonio, line 15: Why `SameSite=Lax`?

validation/jinja2/validation/_segment_card.html
```diff
@@ -33,7 +33,13 @@
                 {% if auth %}
                     <button type="button" class="button button--fail button--flag right" data-toggle="modal" data-target="#{{ phrase.id }}-modal" data-cy="flag-button">Something's wrong with this entry</button>
                 {% endif %}
-                <p class="card__title--cree" data-cy="transcription">{{ phrase.transcription }}</p>
+                <p class="card__title--cree" data-cy="transcription">
+                    {% if request.COOKIES['macron'] == 'on' %}
```
eddieantonio, line 37: If ever you need to support multiple orthographies, I'd take a look at how itwêwina handles this. Ultimately, it's using a cookie to control the orthography, but it has explicit support for multiple orthographies and how to convert to each one.

See: https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/master/CreeDictionary/morphodict/orthography.py

validation/jinja2/validation/_menu.html
```diff
@@ -0,0 +1,35 @@
...
+                {% else %}
+                    <li class="menu__options--right" onclick="showMacron('off')"><a href="#" class="menu__options-link">Show e without macron</a></li>
+                {% endif %}
+                <li class="menu__options--right"><a href="/logout" id="logout-link" class="logout-link">Logout</a></li>
+
+            {% else %}
+                <li><a href="/login" id="login-link" class="login-link">Login</a></li>
+            {% endif %}
+        </ul>
+    </div>
```
eddieantonio: Fun fact: there is a `<menu>` element in the HTML living standard, but it's gone from an old deprecated usage to a new usage (this use case right here). That said, [browers overwhelmingly do not support it](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu#browser_compatibility).

For these kind of menus, I have used [`<details>`/`<summary>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) to enable this functionality without JavaScript.

validation/jinja2/validation/_segment_card.html
```diff
@@ -33,7 +33,13 @@
...
+                        {{ phrase.transcription }}
+                    {% else %}
+                        {{ phrase.transcription | replace("ê", "e") }}
```
eddieantonio, line 40: Pedantic note: the diacritic on the ê [is a circumflex](https://en.wikipedia.org/wiki/Circumflex#Length:~:text=In%20western%20Cree%2C); a macron is [a straight line on top](https://en.wikipedia.org/wiki/Macron_(diacritic)#cite_ref-15:~:text=In%20western%20Cree), like this: ē.

In ALTLab-internal usage, I prefer using the precise names, but I understand that normal folks call both "macrons" — circumflex is such a weird word!

validation/static/menu.js
nienna73: If I don't provide any sort of CORS information, I get a warning saying that the cookie will expire soon and this gets rid of the warning. I'll do some research to see if there's a better option, though.

validation/jinja2/validation/_menu.html
nienna73: Oh beautiful! I'll see if <details>/<summary> will do the trick!

nienna73: Okay, I would edit that, but it's just...too good. The HTML got rendered

validation/jinja2/validation/_segment_card.html
nienna73: This is valid. I was lowkey waiting to see if someone would comment on the wording. I can change it to circumflex and see if we get any user feedback about the wording.

validation/static/menu.js
nienna73: I think this is actually what we want: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#lax

Lax is the new default and makes sure the cookie is passed when following a link to the page.

eddieantonio (approved):
LGTM! Merge it!

PR #186, opened 2021-04-15
Issue modification

nienna73:
## What's in this PR:
* users can suggest the correct speaker code when marking an incorrect speaker
* reworked the wording in the Issue pop-up
* made comment non-mandatory
* refactor both modals (pop-ups) into their own files for easier file management


andrewdotn (approved):
Since we’re not using the HTTP `referer` header sent by the browser, but our own custom form field, maybe spell it correctly?

eddieantonio (approved):
LGTM!

PR #189, opened 2021-04-21
Issue modification

nienna73:
## What's in this PR:

* Made suggestion boxes larger
* Made the modal larger overall
* Added a save button to the header

### DIFF
--- a/validation/static/styles.css
+++ b/validation/static/styles.css
@@ -440,7 +440,7 @@ a:hover {
   color: var(--custom-purple);
 }
 
-textarea {
+.issue__textarea {
   width: 80%;
   height: 3rem;
 }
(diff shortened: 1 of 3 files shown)


validation/static/styles.css
```diff
@@ -439,3 +439,8 @@ a:hover {
 .logout-link:hover, .login-link:hover {
   color: var(--custom-purple);
 }
+
+textarea {
```
eddieantonio: For the least amount of surprises in the future, you may want to make a class and apply it to the specific `<textarea>`(s) you want. That way, if you add `<textarea>`s in the future, you won't be asking --- why is this thinner than I think it should be!

eddieantonio (approved):
LGTM. Just one minor thing.

validation/static/styles.css
nienna73: Absolutely, I completely forgot I could apply styles directly to text areas. Fixed!

PR #190, opened 2021-04-23
Wrong word

nienna73:
## What's in this PR:
* changed the wording for the "report an issue" button
* added a button to mark the wrong word for a recording
* moved the "Wrong Speaker" button into the same column as the "Wrong Word" button
* ~added more options for phrase and recording status/quality~   ~made a Status object that the Recording refers to~ Added options for "wrong word" and "wrong speaker" to the Recording model
* fixed some small bugs relating to the button and header colours reflecting the review status of the word
* pop-ups reload to a specific card, not the top of the page

Here's an example card with one recording in each state:
Screen Shot 2021-04-23 at 11 49 20 AM


validation/jinja2/validation/_base.html
```diff
@@ -54,7 +54,7 @@
   </nav>
 
 
-  <main role="main" class="container">
+  <main role="main" class="main_container">
```
eddieantonio: Incidentally, `<main role="main">` is, not only redundant in writing, but also redundant in meaning. See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#landmark

```suggestion
  <main class="main_container">
```

validation/jinja2/validation/_wrong_buttons.html
```diff
@@ -0,0 +1,42 @@
...
+
+{% macro button_col(recording, csrf_token, request) %}
+
+    {% if recording.quality == "wrong word" %}
+        <button class="button button--neutral wrong-speaker-button" data-cy="wrong-button" data-toggle="modal" data-target="#{{ recording.id }}-modal" data-rec-id={{ recording.id }}>Wrong Speaker</button>
+        <button class="button button--neutral-solid wrong-word-button" data-cy="wrong-button" data-rec-id={{ recording.id }} onclick="showWrongWordDiv('{{ recording.id }}')">Wrong Word</button>
+    {% elif recording.quality == "wrong speaker" %}
+        <button class="button button--neutral-solid wrong-speaker-button" data-cy="wrong-button" data-toggle="modal" data-target="#{{ recording.id }}-modal" data-rec-id={{ recording.id }}>Wrong Speaker</button>
+        <button class="button button--neutral wrong-word-button" data-cy="wrong-button" data-rec-id={{ recording.id }} onclick="showWrongWordDiv('{{ recording.id }}')">Wrong Word</button>
+    {% else %}
```
eddieantonio: What if a recording has both wrong word _and_ the wrong speaker?

validation/jinja2/validation/_wrong_buttons.html
```diff
@@ -0,0 +1,42 @@
...
+      MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+      GNU Affero General Public License for more details.
+
+      You should have received a copy of the GNU Affero General Public License
+      along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#}
+
+{% macro button_col(recording, csrf_token, request) %}
+
+    {% if recording.quality == "wrong word" %}
```
eddieantonio: I feel like duplication here can be simplfied by specifying each button exactly _once_ and toggling the `button--neutral-solid` class in the template.

---

**NOTE**: you do NOT need to implement the following, but I might as well mention it:

 I believe the more approriate markup for this toggle would be to use a `<input type="checkbox">`, wrap it in a `<label>` and use the checkbox's `checked` attribute to toggle the appropriate class on the button. I don't expect this implementation in this PR though!

Unfortunately, [no browser supports the `:has()` pseudo-class](https://caniuse.com/css-has) which would make this CSS possible:

```css
label.wrong-speaker-button:has(> input[type=checkbox][checked]) {
    color: var(--solid-button-text-color);
    background-color: var(--solid-button-background);
}
```

validation/models.py
```diff
@@ -361,6 +362,8 @@ class Recording(models.Model):
     GOOD = "good"
     BAD = "bad"
     UNKNOWN = "unknown"
+    WRONG_SPEAKER = "wrong speaker"
```
eddieantonio: I just realized that the `quality` field should have the `choices` attribute :/

See: https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.Field.choices

validation/static/index.js
```diff
@@ -155,5 +171,24 @@ document.addEventListener('DOMContentLoaded', () => {
...
+
+function hideWrongWordDiv(recordingId) {
+    const wrongWordDivs = document.getElementsByClassName("rec-wrong-word");
+    for (let d of wrongWordDivs) {
+        if (d.dataset.recId === recordingId) {
+            d.classList.add('menu__none')
+            d.classList.remove('menu__block')
+        }
+    }
+}
```
eddieantonio: I'm noticing that this pattern continues to be duplicated:

```javascript
const elements = document.getElementsByClassName(SOME_CLASS_NAME);
for (let e of elements) {
    if (e.dataset.recId === recordingID) {
      e.classList.add(OLD_CLASS);
      e.classList.remove(NEW_CLASS);
    }
}
```

Since there's two things going on here, I suggest this Extract Function refactoring:

```javascript
function* yieldElementWithClassAndRecordingId(className, recordingId) {
  const elements = document.getElementsByClassName(className);
  for (let e of elements) {
      if (e.dataset.recId === recordingID) {
        yield e
      }
  }
}
```

**And you can use `.replace()` instead of `.add()` and `.remove()`**:

```javascript
e.classList.replace(OLD_CLASS_NAME, NEW_CLASS_NAME)
```

This will result in code like:

```suggestion
function showWrongWordDiv(recordingId) {
    for (let d of yieldElementWithClassAndRecordingId("rec-wrong-word", recordingID)) {
        d.classList.replace('menu__none', 'menu__block')
    })
}

function hideWrongWordDiv(recordingId) {
    for (let d of yieldElementWithClassAndRecordingId("rec-wrong-word", recordingID)) {
        div.classList.replace('menu__block', 'menu__none')
    })
}
```

validation/static/styles.css
```diff
@@ -5,6 +5,10 @@ main {
   margin: 1em auto;
 }
 
+.main_container {
+  width: 90%;
```
eddieantonio, line 9: Oh dear! I wonder why this was necessary!

eddieantonio (commented):
I'm a bit concerned about the case when a recording is both has the wrong word AND the wrong speaker. I guess it's worse if it's the wrong word... but still :/

validation/jinja2/validation/_wrong_buttons.html
nienna73: Excellent point. The UI would just show the most recent change and the DB would store that same most recent change. BUT there would be an Issue filed for both cases, so anyone with admin access can go in and see that the recording has both the wrong speaker code and the wrong word. In the future, I would consider making a linguist-specific interface for viewing and resolving these issues.

validation/static/styles.css
nienna73: I wanted the cards to be a little wider so things wouldn't jump around, which meant making the main container larger so the card could be 100% the width of the container. I don't want it to span the full width of the screen, hence 90%.

validation/jinja2/validation/_base.html
nienna73: Ah, good to know. I can take out the `role` part 👍

validation/static/index.js
nienna73: `classList.replace` works on everything except the header. The header is in 1 or 3 states and that state is unknown. If I try to replace a class that isn't being applied to the header, it fails and the header winds up with no style. Otherwise this refactor is awesome!

validation/models.py
```diff
@@ -353,6 +355,16 @@ def generate_primary_key(sender, instance, **kwargs):
 SHA256_HEX_LENGTH = 64
 
 
+class Status(models.Model):
```
eddieantonio: I would actually make this a [OneToOneField](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.OneToOneField) that references the recording HERE, rather than having the recording reference the issue.

i.e.,

```suggestion
class Status(models.Model):
    recording = models.OneToOneField("Recording", on_delete=models.SET_NULL)
```

(then delete the `ForeignKey` in Recording).

This states that a `Status` **must** be associated with exactly one `Recording`. Otherwise, the schema technically allows for "floating" `Status` records that are not associated with any recording.

Django then will do some ORM magic such that accessing `recording.status` either gives you `None` or _exactly_ one `Status` object!

eddieantonio: Addendum: I was wrong about the `on_delete` field; it should be `CASCADE` so that when (if?) a recording gets deleted, its status will be deleted—no stranded data!

```python
class Status(models.Model):
    recording = models.OneToOneField("Recording", on_delete=models.CASCADE)
```

eddieantonio: Addendum addendum:

`recording.status` will give you a `RelatedObjectDoesNotExist` error instead of None :/

validation/jinja2/validation/_segment_card.html
```diff
@@ -18,22 +18,27 @@
 {% import 'validation/_macros.html' as macros %}
 {% import 'validation/_issue_modal.html' as issue_modal %}
 {% import 'validation/_speaker_code_modal.html' as speaker_code_modal %}
+{% import 'validation/_wrong_buttons.html' as wrong_buttons %}
 
 {% macro segment_card(phrase, recordings, speakers, is_linguist, is_expert, auth, forms, csrf_token, request) %}
 
-<article class="card shadow-sm" data-cy="segment-card">
+    <article id='{{ phrase.id }}' class="card shadow-sm" data-cy="segment-card">
```
andrewdotn:
```suggestion
    <article id='phrase-{{ phrase.id }}' class="card shadow-sm" data-cy="segment-card">
```

You’d have to change the references as well, but an HTML id that looks like it’s purely a number might cause confusion down the road.

andrewdotn (commented):
Squashing the migrations is a good step, but it could use a little bit more tweaking.
  - Production is currently on 0006 (in `/opt/speech-db` run `docker-compose exec web ./manage.py showmigrations`), so you should squash 7-15, and then commit only the squashed file for those and not the individual files they came from
  - `makemigrations` takes a `-n` option, and `squashmigrations` takes `--squashed-name` so that you give migrations more meaningful names like “wrong_phrase” or “add_foo_table” or “drop_blah_column”

I can show you a bit more about migrations some time when we are pairing.

andrewdotn (approved):
Only thing I’d check for is whether there’s a potential for conflict between `data-target="#{{ phrase.id }}-modal"` and `data-target="#{{ recording.id }}-modal"`

Other than that, looks good!

nienna73:
`recording.id` is always a hash and `phrase.id` is an incremental number that Django assigns on import. So there's virtually no chance they conflict.

PR #192, opened 2021-04-30
Only show options for new entries

nienna73:
## What's in this PR:
* a new management command for determining if the word is already in MD
* some logic for when to show the validation buttons for a phrase
* updates to the tests and test-db fixture

**Notes:**
The new management command relies on the presence of a file called `treatedMD.csv` in the `private` folder of the project. This file currently only exists on my local machine and it will need to be copied over the Whorf before we can run the command. Fear not, it's just a slightly altered version of a file in the Google Drive, so only having it on my machine isn't nearly as scary as it sounds.


validation/management/commands/findorigin.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python3
+# -*- coding: UTF-8 -*-
+
```
andrewdotn:
```suggestion
```

The first line could be useful if the script had a main method. But it doesn’t.

andrewdotn (approved):

PR #195, opened 2021-05-14
Fix/single quote

nienna73:
## What's in this PR:
* fixed issue where single quote was preventing the edit form from loading

https://user-images.githubusercontent.com/28357720/118331406-d6c93680-b4c5-11eb-964d-fbaaf89c100e.mov


andrewdotn:
I think you might want https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#urlencode ? It’ll turn a string like `foo"ab` into `foo%22ab`.

validation/jinja2/validation/segment_details.html
```diff
@@ -48,7 +48,7 @@
         <button class="button button--success button--back" onclick='goBack()'>Back</button>
         {% if auth %}
             <button data-cy="edit-button" class="button button--fail button--right button--edit"
-                onclick="showDiv('{{ segment_name }}')"
+                onclick='showDiv("{{ segment_name }}")'
```
eddieantonio:
```suggestion
                onclick='showSegmentDiv()'
                data-segment-name="{{ segment_name }}"
```

With the appropriate function defined _something_ like this:

```javascript
function showSegmentDiv() {
    const segmentName = this.target.dataset.segmentName;
    return showDiv(segmentName);
}
```

nienna73:
I tried using `urlencode`, but the reason I need the strings in the first place is partly to get something using the string as part of the id. So if the string has `%`-escaped characters in it, it doesn't work 😢

validation/jinja2/validation/segment_details.html
nienna73: I think this will work

PR #196, opened 2021-05-17
feat: robots.txt that disallows all crawling

eddieantonio:
Googlebot probably should not be crawling the speech-db pages, at least for now.


nienna73 (approved):
Looks good! Thanks for doing this :)

eddieantonio:
@nienna73 this should probably be merged into production soon. Could you please do that?

PR #203, opened 2021-06-02
Styles

nienna73:
## What's in this PR:
* split styles into multiple files

## Why this PR?
The single style sheet was getting very long and confusing--it was hard for me to navigate. Now components have their own files whenever possible. I'm willing to discuss this change if people disagree.


eddieantonio (approved):
LGTM! I like the use of `@import` for this!

PR #204, opened 2021-06-02
feat: searches fuzzy transcription too

nienna73:
## What's in this PR:
* search compares against the fuzzy transcription, too

**Notes:**
Search terms can have **all** the correct accents, or **none** of the accents, nothing in between.


andrewdotn (approved):

validation/views.py
```diff
@@ -136,7 +136,9 @@ def search_phrases(request):
 
     query = request.GET.get("query")
     all_matches = Phrase.objects.filter(
-        Q(transcription__contains=query) | Q(translation__contains=query)
+        Q(transcription__contains=query)
+        | Q(fuzzy_transcription__contains=query)
```
eddieantonio: Whenever you're querying the `fuzzy_transcription` field, use `to_indexable_form()`. Here's an example of its usage: https://github.com/UAlbertaALTLab/recording-validation-interface/blob/7aa44c0a3bfd18863841645ded4fd95598de284b/validation/views.py#L348

```suggestion
        | Q(fuzzy_transcription__contains=to_indexable_form(query))
```

This takes care of the "must have all accents or no accents" problem. Now you can have any amount of accents because `to_indexable_form()` is intended to match things with entries in `fuzzy_transcription`

validation/views.py
```diff
@@ -210,6 +212,7 @@ def advanced_search_results(request):
 
     filter_query = []
     if transcription:
+        filter_query.append(Q(fuzzy_transcription__contains=transcription))
```
```suggestion
        filter_query.append(Q(fuzzy_transcription__contains=to_indexable_form(transcription)))
```

eddieantonio (commented):
Needs more [`to_indexable_form()`](https://github.com/UAlbertaALTLab/recording-validation-interface/blob/7aa44c0a3bfd18863841645ded4fd95598de284b/librecval/normalization.py#L95)!

eddieantonio:
`fuzzy_transcription` is `transcription` but passed through `to_indexable_form()`:

https://github.com/UAlbertaALTLab/recording-validation-interface/blob/7aa44c0a3bfd18863841645ded4fd95598de284b/validation/models.py#L193-L196

validation/views.py
nienna73: Awesome, thank you!

eddieantonio (approved):
LGTM!

PR #205, opened 2021-06-03
Approve linguist

nienna73:
## What's in this PR:
* new expert or linguist sign up sends an email to all admins

**Notes:**
Admins are manually declared in `settings.py`
We need to configure settings on the server before this will work! (I did this already, but still good to note)


andrewdotn (approved):
Looks good

Just to confirm: the email account configured on the server is one specifically created to be used only for sending these emails, and nothing bad or embarrassing could happen if someone else was able to log in to it?

nienna73:
Yes, the account is just for sending emails, no one actually checks it or uses it personally from what I understand.

PR #206, opened 2021-06-04
Jump to Page

nienna73:
## What's in this PR:
* added a "Jump to page" field in the navigation
* added Cypress tests for navigation


validation/jinja2/validation/_macros.html
```diff
@@ -57,30 +23,33 @@
...
+                Next
+            </a></li>
+            <li class="nav-item"><a class="nav__link" data-cy="nav-last" href={{ encode_query_with_page(query, page=page.paginator.num_pages) }}>
+                Last
+            </a></li>
+        {% endif %}
+
+    </ul>
+    <div class="center">
+        Jump to Page: <input type="text" data-cy="page-num-input" class="input--short"> <button type="button" data-cy="go-button-nav" class="button button--success button--small">Go</button>
```
eddieantonio, line 51: I'm not 100% sure this has the same intended effect but this could be replaced by a form (zero JavaScript):

```suggestion
        <form method="GET">
            <label> Jump to Page: <input type="number" name="page" data-cy="page-num-input" class="input--short"></label>
            <button type="submit" data-cy="go-button-nav" class="button button--success button--small">Go</button>
        </form>
```

eddieantonio (approved):
LGTM, but it _might_ be possible to get rid of the JavaScript by using a `<form>`.

validation/jinja2/validation/_macros.html
nienna73: I tried using a form initially and couldn't quite figure it out. Consider that `page=` often isn't the only parameter. I would also likely pull the form out of `macros` since it needs to `csrf_token`. I'm open to it if it works though!

eddieantonio: Oh yeah, you're right! This will destroy any and all other params other than page! Note, `<form method="GET">` does not requrire a CSRF token; [only `POST` (and other, hipster methods) require the CSRF token ](https://docs.djangoproject.com/en/3.2/ref/csrf/#how-to-use-it).

Okay, keep the JS! 😂

andrewdotn: *do not actually do the following*, I only bring it up because a puzzle was implicitly posed:

Make it work by adding `<input type="hidden">` on the server side for every existing GET param?

validation/static/nav.js
```diff
@@ -0,0 +1,16 @@
...
+document.addEventListener("DOMContentLoaded", () => {
+    for (let b of document.querySelectorAll('[data-cy="go-button-nav"]')) {
+        b.addEventListener("click", (e) => {
+            let pageNo = document.querySelector('[data-cy="page-num-input"]').value;
+            if (!pageNo) {
+                return;
+            }
+
+            let searchParams = new URLSearchParams(window.location.search);
+            searchParams.set("page", pageNo)
```
andrewdotn, line 12: So much better!

andrewdotn (approved):

PR #207, opened 2021-06-04
Findorigin

nienna73:
## What's in this PR:
* when the word already exists in MD, also save the translation (not just the origin)


validation/management/commands/findorigin.py
```diff
@@ -33,27 +33,27 @@ def handle(self, *args, **options):
...
-            if not phrase.origin:
-                normalized = normalize_sro(phrase.transcription)
-                normalized = remove_diacritics(normalized)
-                if normalized in md_words:
-                    phrase.origin = "MD"
-                else:
-                    phrase.origin = "new"
-                phrase.save()
+            normalized = normalize_sro(phrase.transcription)
+            normalized = remove_diacritics(normalized)
```
eddieantonio, line 37: Is it for sure that a term in the Maskwacîs dictionary will be written in normalized SRO, without diacritics?

validation/management/commands/findorigin.py
```diff
@@ -33,27 +33,27 @@ def handle(self, *args, **options):
...
+    :return: a dict of Cree words and their translations
     """
-    words = []
+    words = {}
     with open("private/treatedMD.csv", "r") as f:
         reader = csv.DictReader(f)
 
         for row in reader:
-            words.append(row["SRO"])
+            words[row["SRO"]] = row["MeaningInEnglish"]
```
eddieantonio, line 56: I'm not sure what file `treatedMD.csv` is, but are we sure that each `row["SRO"]` is unique? What happens if multiple rows have the same `row["SRO"]`?

eddieantonio (approved):
LGTM\*, provided that the questions I asked are non-issues.

validation/management/commands/findorigin.py
eddieantonio, line 56: I guess to answer my own question, you could have this:

```suggestion
            assert row["SRO"] == remove_diacritics(normalize_sro(row["SRO"]))
            words[row["SRO"]] = row["MeaningInEnglish"]
```

Or, if this is okay:

```suggestion
            normalized = remove_diacritics(normalize_sro(row["SRO"]))
            words[normalized] = row["MeaningInEnglish"]
```

validation/management/commands/findorigin.py
nienna73: The words in the Maskwacîs dictionary have been pre-treated to not contain diacritics to make matching easier. I understand this means I could run into some homograph ambiguities, but I think those can be manually resolved in the interface.

validation/management/commands/findorigin.py
nienna73: I'm 95% confident that each `row["SRO"]` is unique, and if it isn't then the later translation will overwrite any earlier ones

PR #220, opened 2021-06-23
Changes to linguist view

nienna73:
## What's in this PR:
* all users can see auto-validated entries
* remove the check boxes from the Issue pop-up
* don't preload translations
* always show edit form when viewing the details of an entry


eddieantonio (approved):
Looks G to me!

PR #227, opened 2021-07-09
fix: remove boolean fields from issue model

nienna73:
## What's in this PR:
* remove the boolean fields from the issue model

I forgot to remove this earlier when I removed it from the pop-up, so it was causing 500 errors 🙁


validation/views.py
```diff
@@ -630,7 +630,6 @@ def prep_phrase_data(request, phrases):
 
 def save_issue(data, user):
     phrase_id = data["phrase_id"]
-    issues = data["issues"]
```
eddieantonio, line 633: Was _this_ what was causing a 500 error (due to a `KeyError`?)

eddieantonio (approved):
LGTM!

validation/views.py
nienna73: It sure was

PR #229, opened 2021-07-13
Api/bulk

nienna73:
## What's in this PR:
* implement bulk search API to return many exactly matched wordforms at once

**Notes:**
There's some refactoring that could be done. Much of the code is duplicated between `search_recordings` and `bulk_search_recordings`, I just can't wrap my head around how to do the refactor at the moment.


validation/tests/test_bulk_recordings_search.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    phrase = Phrase.objects.filter(transcription=query).first()
+    speaker = Speaker.objects.filter(code=speaker_code).first()
+
+    url = reverse("validation:bulk_search_recordings")
+
+    response = client.get(url + "?" + urllib.parse.urlencode([("q", query)]))
+
+    assert "Access-Control-Allow-Origin" in response, "Missing requried CORS headers"
```
andrewdotn:
```suggestion
    assert "Access-Control-Allow-Origin" in response, "Missing required CORS headers"
```

validation/views.py
```diff
@@ -381,6 +381,64 @@ def make_absolute_uri_for_speaker(code: str) -> str:
...
+    """
+
+    def make_absolute_uri_for_recording(rec: Recording) -> str:
+        uri = rec.compressed_audio.url
+        if uri.startswith("/"):
+            # It's a relative URI: build an absolute URI:
+            return request.build_absolute_uri(uri)
+
+        # It's an absolute URI already:
+        assert uri.startswith("http")
```
andrewdotn: I hope it’s ok to be a little bit nitpicky here, but in case some random filename like `http_stuff.txt` starts with `http` someday:

```suggestion
        assert uri.startswith("http://") or uri.startswith("https://")
```

.github/workflows/test.yml
```diff
@@ -37,15 +37,6 @@ jobs:
...
       - run: pip install pipenv
 
-      # Copy-paste from
-      # https://github.com/actions/cache/blob/main/examples.md#python---pipenv
-      - uses: actions/cache@v2
-        with:
-          path: ~/.local/share/virtualenvs
-          key: ${{ runner.os }}-pipenv-${{ hashFiles('Pipfile.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-pipenv-
```
eddieantonio: Restoring the cache appears to be broken today. If this is present, the `python` executable disappears from the rest of the workflow, which is... a bit of a problem.

eddieantonio: Fixed in #233

recvalsite/settings.py
```diff
@@ -106,10 +106,14 @@
     "RECVAL_SQLITE_DB_PATH", default=BASE_DIR / "db.sqlite3", cast=Path
 )
 
+USE_TEST_DB = config("USE_TEST_DB", default=False)
+
```
andrewdotn: Normally Django’s unit tests use a temporary in-memory database that is quite fast. We only `USE_TEST_DB` in morphodict because the import process is too slow to load the required test data for every test. And it’s a bit of a pain.

If there isn’t a specific need to have a bunch of data that’s time-consuming to compute in the test database, I would recommend entirely skipping anything to do with `USE_TEST_DB`.

(If this is meant for cypress, I have other suggestions for that, but I don’t think that’s in scope for this PR.)

```suggestion
```

recvalsite/settings.py
```diff
@@ -106,10 +106,14 @@
...
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.sqlite3",
-        "NAME": os.fspath(RECVAL_SQLITE_DB_PATH),
+        "NAME": os.fspath(BASE_DIR / "test_db.sqlite")
+        if USE_TEST_DB
+        else os.fspath(RECVAL_SQLITE_DB_PATH),
```
andrewdotn:
```suggestion
        "NAME": os.fspath(RECVAL_SQLITE_DB_PATH),
```

validation/tests/test_bulk_recordings_search.py
```diff
@@ -0,0 +1,106 @@
+"""
+Unit/Integration tests for the bulk recordings search API.
+"""
+import urllib
+
+import pytest  # type: ignore
+from django.core.management import call_command
+from django.shortcuts import reverse  # type: ignore
```
andrewdotn, line 8: If mypy is complaining about these lines, I’d add ignores to `mypy.ini` so they apply for all imports of these packages, instead of having to remember to add `# type: ignore` every time they are imported. There are examples of how to ignore packages in the morphodict `mypy.ini`.

```suggestion
import pytest
from django.core.management import call_command
from django.shortcuts import reverse
```

validation/views.py
```diff
@@ -381,6 +381,64 @@ def make_absolute_uri_for_speaker(code: str) -> str:
...
+    for term in query_terms:
+        # Assume the query is an SRO transcription; prepare it for a fuzzy match.
+        result_set = Recording.objects.filter(
+            phrase__transcription=term,
+            speaker__gender__isnull=False,
+        )
+        # No bad recordings!
+        result_set = result_set.exclude(quality=Recording.BAD)
+        result_set = result_set.exclude(wrong_word=True)
+        result_set = result_set.exclude(wrong_speaker=True)
```
andrewdotn: Totally optional, but if you do `result_set = Recordings.objects.filter(…).exclude(quality=Recording.BAD).exclude(wrong_word=True).exclude(wrong_speaker=True)` and let black format it, you’ll be able to say `result_set` six fewer times.

andrewdotn (commented):
Looks good overall, just a few minor changes requested.

recvalsite/settings.py
nienna73: `USE_TEST_DB` was initially included so we could load specific entries into the DB and look for them in the tests. We don't actually use it in the tests anymore, but it seemed like a fine variable to have.

eddieantonio: On second thought, introducing a new environment variable that is unused and undocumented creates a future maintance problem; i.e., somebody will ask "what is this here for?" "why did they use this?" "where did they use this?" "what did this affect?".

Plus, `USE_TEST_DB` is not really justified in the unit tests for this project anyway, since we learned that populating the data with fixtures is _really_ quick an snappy!

So I'm just going to pretend it never got introduced...

nienna73: Fair enough, it doesn't really make sense to have something in there "just in case".

PR #231, opened 2021-07-14
Fix wrong word button

nienna73:


eddieantonio (approved):
Are there any other places in the codebase that call `Issue()` or may exhibit the same problem? I worried we might be playing bug whack-a-mole here...

PR #234, opened 2021-07-16
refactor: clean up recordings search API endpoints

eddieantonio:
Just various attempts at try to remove the duplication in the two various recording search API endpoints. A lot of it is gone, but there is some duplication that I think is okay to have (or, more precisely, the solution I'm thinking of to get rid of the duplication is just way too much >.<).

**EDIT**: this is what I mean by "some duplication is okay to have": https://github.com/UAlbertaALTLab/recording-validation-interface/compare/refactor/clean-up-recordings-search...DO-NOT-MERGE/template-method-refactor-i-tried?expand=1

### DIFF
--- a/validation/views.py
+++ b/validation/views.py
@@ -337,14 +337,13 @@ def search_recordings(request, query):
     for form in word_forms:
         # Assume the query is an SRO transcription; prepare it for a fuzzy match.
         fuzzy_transcription = to_indexable_form(form)
-        result_set = Recording.objects.filter(
+        all_matches = Recording.objects.filter(
             phrase__fuzzy_transcription=fuzzy_transcription,
         )
-        # No bad recordings!
-        result_set = exclude_known_bad_recordings(result_set)
+        results = exclude_known_bad_recordings(all_matches)
 
         recordings.extend(
-            create_recording_result_json(request, recording) for recording in result_set
+            create_recording_result_json(request, recording) for recording in results
         )
 
     response = JsonResponse(recordings, safe=False)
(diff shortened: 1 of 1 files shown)


validation/views.py
```diff
@@ -699,3 +644,60 @@ def save_issue(data, user):
...
+    Returns an absolute URL for the compressed audio recording.
+    This can be directly used in an <audio> tag to hear the recording on a webpage!
+    """
+    uri = rec.compressed_audio.url
+    if uri.startswith("/"):
+        # It's a relative URI: build an absolute URI:
+        return request.build_absolute_uri(uri)
+
+    # It's an absolute URI already:
+    assert uri.startswith(("http://", "https://"))
```
andrewdotn, line 685: Cool, didn’t know you could do this.

validation/views.py
```diff
@@ -699,3 +644,60 @@ def save_issue(data, user):
     )
 
     new_issue.save()
+
+
+def create_recording_result_json(request: HttpRequest, rec: Recording):
```
andrewdotn, line 649: Maybe make this an instance method, `Recording.serialize()` ?

validation/views.py
```diff
@@ -699,3 +644,60 @@ def save_issue(data, user):
...
+    if uri.startswith("/"):
+        # It's a relative URI: build an absolute URI:
+        return request.build_absolute_uri(uri)
+
+    # It's an absolute URI already:
+    assert uri.startswith(("http://", "https://"))
+    return uri
+
+
+def exclude_known_bad_recordings(recordings: QuerySet):
```
andrewdotn, line 689: This *could* be a [Manager](https://docs.djangoproject.com/en/3.2/topics/db/managers/), e.g., `Recording.not_known_bad_objects` in addition to `Recording.objects`

andrewdotn (approved):

nienna73 (approved):
Thank you!

PR #235, opened 2021-07-16
refactor: additional search refactorings

eddieantonio:
# What's in this PR?

Implemented refactoring suggestions from the review comments here: https://github.com/UAlbertaALTLab/recording-validation-interface/pull/234#pullrequestreview-708607824

Each change is an individual commit, so it's easy to `git cherry-pick` a subset of refactors, if desired.

### DIFF
--- a/validation/models.py
+++ b/validation/models.py
@@ -23,7 +23,6 @@ from pathlib import Path
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import QuerySet
 from django.db.models.signals import pre_save
 from django.dispatch import receiver
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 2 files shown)


validation/models.py
```diff
@@ -355,6 +364,26 @@ def generate_primary_key(sender, instance, **kwargs):
...
+            .exclude(wrong_word=True)
+            .exclude(wrong_speaker=True)
+            # We use the "gender" field as a proxy to see whether a speaker's data has
+            # been properly filled out: exclude speakers whose gender field has not been
+            # input. An admin must put *something* in the gender field before the
+            # speaker shows up in API results.
+            .exclude(speaker__gender__isnull=True)
+        )
+
+
```
nienna73: I'm okay with having this one in `views.py` as that's where I do a lot of other query filtering and I find it easier to have it all in one file. I do understand how this could be cleaner though.

andrewdotn (approved):

eddieantonio:
@nienna73 I reverted the extra manager according to your comment. I think having the function `exclude_known_bad_recordings()` pretty much explains what it's supposed to do!

nienna73 (approved):

PR #240, opened 2021-07-22
View issues

nienna73:
## What's in this PR:
* a view to look at and resolve open issues
* tests for this functionality

## More about this PR:
Resolving an issue relating to a phrase changes the translation and transcription for that phrase if new ones have been provided.
Resolving an issue relating to a recording can do a few things:
1. Changing the speaker simply associates a new speaker to the recording
2. Changing the word in the recording will first look for an entry with the same exact spelling as the new word or phrase. If one exists, it will associate this recording with this existing phrase. If one does not exist, one will be created using the relevant information from the recording.


validation/forms.py
```diff
@@ -111,3 +111,35 @@ class FlagSegment(forms.ModelForm):
     class Meta:
         model = Issue
         fields = ["phrase_id"]
+
+
+class EditIssueWithRecording(forms.ModelForm):
+    phrase = forms.CharField(
+        widget=forms.Textarea(),
+        required=False,
+        help_text="The sentence or phrase is this recording is actually...",
```
eddieantonio:
```suggestion
        help_text="What word or phrase this recording is actually for",
```

validation/forms.py
```diff
@@ -111,3 +111,35 @@ class FlagSegment(forms.ModelForm):
...
+
+    translation = forms.CharField(
+        max_length=200,
+        required=False,
+    )
+
+    class Meta:
+        model = Phrase
+        fields = ["transcription", "translation"]
+        help_texts = {"transcription": "\n", "translation": "\n"}
```
eddieantonio: Why is the `help_texts` for these (and `EditIssueWithRecording`) a newline?

validation/migrations/0009_issue_status.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.20 on 2021-05-11 15:44
```
eddieantonio: How are migrations initiated on the server?

validation/models.py
```diff
@@ -451,6 +451,17 @@ class Issue(models.Model):
...
 
+    RESOLVED = "resolved"
+    OPEN = "open"
+    STATUS_CHOICES = [(RESOLVED, "Resolved"), (OPEN, "Open")]
+
+    status = models.CharField(
+        help_text="The status of the issue",
+        max_length=64,
+        choices=STATUS_CHOICES,
+        default=OPEN,
```
eddieantonio, line 531: :+1:

validation/static/styles/button_styles.css
```diff
@@ -37,6 +37,15 @@
   border: 2px solid var(--custom-orange);
```
eddieantonio, line 37: Why was `validation/static/styles.css` created?

(GitHub won't let me comment on an empty file 😖)

validation/views.py
```diff
@@ -484,9 +492,99 @@ def register(request):
...
+
+def view_issue_detail(request, issue_id):
+    issue = Issue.objects.filter(id=issue_id).first()
+
+    form = None
+    if issue.recording:
+        form = EditIssueWithRecording(
+            request.POST, initial={"rec_id": str(issue.recording.id)}
+        )
+        rec = Recording.objects.filter(id=issue.recording.id).first()
```
```suggestion
        rec = Recording.objects.get(id=issue.recording.id)
```

validation/views.py
```diff
@@ -484,9 +492,99 @@ def register(request):
...
+        phrase = None
+    if issue.phrase:
+        form = EditIssueWithPhrase(request.POST, initial={"phrase_id": issue.phrase.id})
+        phrase = Phrase.objects.filter(id=issue.phrase.id).first()
```
```suggestion
        phrase = Phrase.objects.get(id=issue.phrase.id)
```

validation/views.py
```diff
@@ -484,9 +492,99 @@ def register(request):
...
+        rec = None
+
+    if request.method == "POST":
+        if form.is_valid():
```
eddieantonio: Is there an `else` for this? i.e, when the `form` is **not** valid? Should this respond with a `4xx` error?

validation/views.py
```diff
@@ -484,9 +492,99 @@ def register(request):
...
+            if rec:
+                speaker_code = form.cleaned_data["speaker"]
+                if speaker_code:
+                    speaker = Speaker.objects.filter(code=speaker_code).first()
```
```suggestion
                    speaker = Speaker.objects.get(code=speaker_code)
```

validation/views.py
```diff
@@ -484,9 +492,99 @@ def register(request):
...
+        is_linguist=user_is_linguist(request.user),
+    )
+    return render(request, "validation/view_issues.html", context)
+
+
+def view_issue_detail(request, issue_id):
+    issue = Issue.objects.filter(id=issue_id).first()
+
+    form = None
+    if issue.recording:
```
eddieantonio, line 518: If you want to make this a bit cleaner, I suggest making two new functions: one that handles `issue.recording` and one that handlers `issue.phrase`. It seems like the big `if`/`elif` do not share a lot of code, so you could totally split them up into separate functions without as many indents!

validation/views.py
```diff
@@ -484,9 +492,99 @@ def register(request):
...
+        form=form,
+        auth=request.user.is_authenticated,
+        is_linguist=user_is_linguist(request.user),
+    )
+    return render(request, "validation/view_issue_detail.html", context)
+
+
+def close_issue(request, issue_id):
+    issue = Issue.objects.filter(id=issue_id).first()
+    issue.status = "Resolved"
```
eddieantonio: ⚠️ there's a minor bug here!  `"Resolved"` is not a valid option but `"resolved"` _is_! It's less error-prone to use the constant instead:

```suggestion
    issue.status = Issue.RESOLVED
```

validation/views.py
```diff
@@ -484,9 +492,99 @@ def register(request):
...
+                        new_phrase = Phrase(
+                            field_transcription=new_word,
+                            transcription=new_word,
+                            translation="",
+                            kind="Sentence" if " " in new_word else "Word",
+                            date=datetime.datetime.now(),
+                            modifier=str(request.user),
+                        )
+                        new_phrase.save()
+                    rec.phrase_id = new_phrase.id
```
eddieantonio: I will be so happy when this line of code is merged in production! 🤩

eddieantonio (commented):
There's a minor bug here (not sure if Django does any checking of its `CharField`'s choices). There's also an empty file added that is unused and I'm not sure if that's intentional.

validation/forms.py
nienna73: This one doesn't need it anymore, you're right. It was necessary before I added the translation and transcription fields above.

For the speaker, though, it's quite necessary for formatting:
With the `\n`:
Screen Shot 2021-07-24 at 1 17 35 PM

Without the `\n`:
Screen Shot 2021-07-24 at 1 17 18 PM


Though there might be a better way to do this using CSS

validation/migrations/0009_issue_status.py
nienna73: They aren't! I have to go in and do them manually every time I release to `production`. I should absolutely automate this process, though.

validation/static/styles/button_styles.css
nienna73: Ah I think there was a merge conflict in the styles and I somehow kept the empty file while resolving it. Thanks for catching this!

validation/views.py
nienna73: Interesting. None of the other forms in `views.py` have an `else` clause to them, but maybe the should 🤔

validation/views.py
nienna73: Thank you!!

nienna73:
This PR officially has an interesting issue preventing me from merging it!

If someone corrects an issue to have a different/new transcription, that transcription can only contain these characters: ptkcshmn yw rl eêiîoôaâ -

If it contains anything other than those, the site crashes due to an assertion error in the Django model (one that we wrote and can change if we would like).

However! We have some entries in the database that use those characters, most of which are proper names, such as:

Brian ekwetaweyimat Marywa. : https://speech-db.altlab.app/segment/16528
môcâya apisînîs niya ewîcikâpawistawak Shaq: https://speech-db.altlab.app/segment/12508
newewepisokanet Elvis: https://speech-db.altlab.app/segment/419

It is unclear to me how these made it into the DB without the DB crying, but they are there.

So, do we proceed with only allowing the "ptkcshmn yw rl eêiîoôaâ -" characters?
Or should I remove that constraint on the model to allow all characters in the English alphabet?

eddieantonio:
> If it contains anything other than those, the site crashes due to an assertion error in the Django model (one that we wrote and can change if we would like).
>
> Brian ekwetaweyimat Marywa. : https://speech-db.altlab.app/segment/16528
> môcâya apisînîs niya ewîcikâpawistawak Shaq: https://speech-db.altlab.app/segment/12508
> newewepisokanet Elvis: https://speech-db.altlab.app/segment/419

:joy: :joy: :joy:

> It is unclear to me how these made it into the DB without the DB crying, but they are there.

The assertion is in the `clean()` method of the model:

https://github.com/UAlbertaALTLab/recording-validation-interface/blob/e7e1b8cac9b14a442d2d563b24863e662a479427/validation/models.py#L190-L191

_Something_ calls `.clean()` — I'm not sure what. Could you please check the traceback to see if the call is coming from within Django's code?

> So, do we proceed with only allowing the "ptkcshmn yw rl eêiîoôaâ -" characters?
> Or should I remove that constraint on the model to allow all characters in the English alphabet?

We need to be careful about this since we've switched to _exact_ matching. I think we can relax (or skip) the constraint for **sentences** but **words** still must follow the strict SRO constraint. 

Aside: apparently the more cromulent thing to do is to to raise a `django.core.expeptions.ValidationError`: https://docs.djangoproject.com/en/3.2/ref/models/instances/#django.db.models.Model.clean

eddieantonio:
@nienna73 I'mma write a quick PR to explain what I mean...

nienna73:
I am raising a `ValidationError`. The `form.is_valid()` calls `clean()`, but I have it explicitly cleaning the transcription as well. In addition to the `ValidationError` and the `clean()` function in `forms.py`, there's another `clean()` function in `models.py` and that's what checks for the allowed characters. I can make a different `clean` function that allows any characters in sentences but only those characters in words.

The form itself is currently a `ModelForm`, which is why the model's `clean()` function gets called. I don't think it needs to be a `ModelForm` though, so I can make it a regular form and do my own validation on it.

eddieantonio:
@nienna73 here's my proposal to change the `Phrase` validation: https://github.com/UAlbertaALTLab/recording-validation-interface/pull/244

I believe model validation is probably what should be used in most circumstances!

validation/models.py
```diff
@@ -183,18 +183,30 @@ def clean(self):
...
     def save(self, *args, **kwargs):
         # Make sure the fuzzy match is always up to date
         self.fuzzy_transcription = to_indexable_form(self.transcription)
         super().save(*args, **kwargs)
 
+    def transcription_is_in_strict_sro(self) -> bool:
+        """
+        Returns True when the transcription is written in perfect SRO.
+        """
+        return self.ALLOWED_TRANSCRIPTION_CHARACTERS.issuperset(self.transcription)
```
andrewdotn, line 208: Cool, did not know about `issuperset()`, or how it can be used so handily here!

andrewdotn (approved):
Looks good! Just be sure to accept [this suggestion](https://github.com/UAlbertaALTLab/recording-validation-interface/pull/240/files#r675207069) before merging.

PR #242, opened 2021-07-24
Feat/suggestion origin

nienna73:
## What's in this PR:
* show the origin of a suggestion that comes from itwêwina
* use `get` instead of `filter` to get a single entry from the DB
* strip incoming text of blank space(s) at the beginning/end

Looks like:
Screen Shot 2021-07-24 at 1 13 23 PM

### DIFF
--- a/validation/helpers.py
+++ b/validation/helpers.py
@@ -179,20 +179,29 @@ def get_translations(results):
     matches = []
 
     for i in results["results"]:
-        translations, analysis = extract_translations(i["lemma_wordform"])
-        if {"translations": translations, "analysis": analysis} not in matches:
-            matches.append({"translations": translations, "analysis": analysis})
+        translations, analysis, sources = extract_translations(i["lemma_wordform"])
+        for (j, translation) in enumerate(translations):
+            source = sources[j]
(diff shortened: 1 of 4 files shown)


validation/helpers.py
```diff
@@ -179,20 +179,29 @@ def get_translations(results):
     matches = []
 
     for i in results["results"]:
-        translations, analysis = extract_translations(i["lemma_wordform"])
-        if {"translations": translations, "analysis": analysis} not in matches:
-            matches.append({"translations": translations, "analysis": analysis})
+        translations, analysis, sources = extract_translations(i["lemma_wordform"])
+        for (j, translation) in enumerate(translations):
+            source = sources[j]
```
eddieantonio: Are `translations` and `sources` parallel arrays? In that case, `zip()` _might_ be cleaner:

```suggestion
        for translation, source in zip(translations, sources):
```

eddieantonio (approved):
LGTM!

validation/helpers.py
nienna73: Nice, thanks!

PR #243, opened 2021-07-24
feat: search for just sentences or just words

nienna73:
## What's in this PR:
* add the option to search for just sentences or just words in the advanced search


eddieantonio (approved):
LGTM!

PR #245, opened 2021-07-30
Speakers

nienna73:
## What's in this PR:
* add speakers page
* extract speaker bios


validation/jinja2/validation/speaker_view.html
```diff
@@ -0,0 +1,29 @@
...
+    </div>
+
+    <div class="speaker-bio">
+        {% if speaker.eng_bio_text %}
+            <h3>Listen in English:</h3>
+            <p>{{ macros.bio_player(speaker.eng_bio_audio) }}</p>
+            {% if request.COOKIES['macron'] == 'on' %}
+                <p>{{ speaker.eng_bio_text }}</p>
+            {% else %}
+                <p>{{ speaker.eng_bio_text | replace("ê", "e") }}</p>
```
andrewdotn, line 17: It could definitely go into a separate PR, but you probably want to create a custom tag here. morphodict has one called ['orth'](https://github.com/UAlbertaALTLab/cree-intelligent-dictionary/blob/cf94e18ac2b6a4be84d6d572f2dd1bd00885f7fd/src/CreeDictionary/morphodict/templatetags/morphodict_orth.py#L13-L33); then you’d just write `{% macron speaker.eng_bio_text %}` or `{% macron "nêhiyawêwin" %}` and the tag function would automatically look at `request.COOKIES` for you, to avoid a bunch of `{% if %} … {% else %}`ing in the templates.

andrewdotn (approved):
My high-level comment here is that, if time permits, I’d look into using [`ImageField`](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ImageField) to store the speaker images. It’s a lot like the FileField that you’re already using, and down the road it’d make some things like, say, generating thumbnails, a lot easier.

validation/jinja2/validation/speaker_view.html
nienna73: I'll definitely look into this! This project needs a fair bit of cleanup in general, so I'll add this to the list of things to fix.

PR #247, opened 2021-08-11
Record audio

nienna73:
## What's in this PR:
* add view to record audio directly into the speech-db

## How it works:
On the frontend:
* login
* navigate to /secrets/record_audio
* add the transcription and translation
* record your audio
* save the audio

On the backend:
* if an entry with that transcription exists, add the recording to that entry
* otherwise create a new phrase entry
* create a new speaker for the logged in user if they aren't already a speaker
* save the recording to the DB
* only display user-submitted recordings to linguists and upon explicit request (filter by user-submitted)

**Notes:**
I want to deploy this to production once its approved so I can see if the speech-db even has permission to save audio files to disk on the server. Once I know that works, I can remove it from `/secrets` and add a link to the page in the header and/or footer.


validation/jinja2/validation/record_audio.html
```diff
@@ -0,0 +1,28 @@
...
+    <p>Use this page to record or re-record entries in Plains Cree. Start by entering the transcription (in Cree)
+    and the translation (in English) for the word or sentence you are going to record. Try to spell the Cree word(s)
+        as correctly as possible, as these recordings will be added to an existing entry if one exists.</p>
+    <p>Next, click the "Record" button. It will likely ask you permission to use your microphone, please click "Allow". You are now recording, as indicated
+    by the word "Recording..." on the button. When you are finished, press "Stop". Listen to your recording. If you're happy
+    with how it sounds, click "Save" next to that recording. Please record yourself saying this word or phrase up to 3 times,
+    clicking save next to each good recording and ignoring any that you are unhappy with. To record a new word or phrase,
+    click the "Record a new Entry" button.</p>
+    <p>Please note that pressing the "Record a new Entry" button or reloading the page will cause all your current unsaved
+        recordings to disappear, so make sure you're done with them and have saved all the good ones before doing either of those actions.</p>
```
andrewdotn, line 16: If you want to later do one of those popups that says, “You have created recordings that you have not uploaded, do you want to continue?”, I think [`beforeunload`](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) is the thing for that.

That also suggests having a *delete* button next to each recording which feels even more feature-creepy, so these are totally just suggestions for the future.

Please remember we’re all telling you all these things you can do with this because we’re *excited* about the *great new feature*!

validation/jinja2/validation/record_audio.html
```diff
@@ -0,0 +1,28 @@
...
+    {{ form }}
+    <div id="controls" class="space">
+        <button id="recordButton" class="button button--success">Record</button>
+{#        <button id="pauseButton" class="button button--neutral" disabled>Pause</button>#}
+        <button id="stopButton" class="button button--fail" disabled>Stop</button>
+    </div>
+    <h3>Recordings</h3>
+    <ol id="recordingsList"></ol>
+    <!-- inserting these scripts at the end to be able to use all the elements in the DOM -->
+    <script type="module" src="https://cdn.rawgit.com/mattdiamond/Recorderjs/08e7abd9/dist/recorder.js"></script>
```
andrewdotn: rawgit.com:
> RawGit is now in a sunset phase and will soon shut down. It's been a fun five
> years, but all things must end.
>
> 
>
> GitHub repositories that served content through RawGit within the  last month
> will continue to be served until at least October of 2019.  URLs for other
> repositories are no longer being served.
>
> 
>
> If you're currently using RawGit, please stop using it as soon as you can.

```suggestion
    <script type="module" src="https://cdn.jsdelivr.net/gh/mattdiamond/Recorderjs@08e7abd99739be6946f19f6806ccb368138f2dd3/dist/recorder.js"></script>
```
If you want to set up bundling for recorderjs so it just comes from npm, I’d be happy to do that in a pairing session.

validation/migrations/0011_auto_20210810_1611.py
```diff
@@ -0,0 +1,76 @@
+# Generated by Django 2.2.24 on 2021-08-10 22:11
```
andrewdotn: These migrations have auto names

validation/static/record.js
```diff
@@ -0,0 +1,125 @@
+"use strict";
+
+// referencing: https://blog.addpipe.com/using-recorder-js-to-capture-wav-audio-in-your-html5-web-site/
+
+let gumStream;      // stream from getUserMedia()
+let rec;            // Recorder.js object
+let input;          // MediaStreamAudioSourceNode
+const AudioContext = window.AudioContext;
+let audioContext = new AudioContext;
```
andrewdotn: Does it still work if you do:

```suggestion
const audioContext = new AudioContext();
```

?

validation/static/record.js
```diff
@@ -0,0 +1,125 @@
...
+let recordButton = document.getElementById("recordButton");
+let stopButton = document.getElementById("stopButton");
+// TODO: do we want a pause button??
+// let pauseButton = document.getElementById("pauseButton");
```
andrewdotn:
```suggestion
```

And likewise with all the other commented-out code in this file.

validation/static/record.js
```diff
@@ -0,0 +1,125 @@
...
+let recordingsList = document.getElementById("recordingsList");
+let recNewEntryButton = document.getElementById("recNewEntryButton");
+
+recordButton.addEventListener("click", startRecording)
+stopButton.addEventListener("click", stopRecording)
+// pauseButton.addEventListener("click", pauseRecording)
```
andrewdotn:
```suggestion
```

validation/views.py
```diff
@@ -707,6 +734,59 @@ def save_wrong_word(request, recording_id):
...
+                date=datetime.datetime.now(),
+                modifier=request.user,
+            )
+            phrase.save()
+
+        speaker = Speaker.objects.filter(user=request.user).first()
+        if not speaker:
+            speaker = Speaker(
+                full_name=request.user.first_name + " " + request.user.last_name,
+                code=request.user.first_name[:3].upper(),
```
andrewdotn: The codes are already **super**-ambiguous and confusing. Automatically adding more is inviting disaster. What will happen if someone else named Brian signs up??

Could we use the user’s login id, e.g., `andrew`, or `U${user.id}` like `U15` for user-submitted? That’ll at least keep it at three characters for the first 100 people.

validation/views.py
```diff
@@ -707,6 +734,59 @@ def save_wrong_word(request, recording_id):
...
+            compressed_audio=audio_data,
+            speaker=speaker,
+            phrase=phrase,
+            timestamp=0,
+            comment=f"Uploaded by user: {request.user}",
+            is_user_submitted=True,
+        )
+        rec.save()
+        source = settings.RECVAL_AUDIO_PREFIX + _id + ".wav"
+        dest = settings.RECVAL_AUDIO_PREFIX + _id + ".m4a"
```
andrewdotn: I find the variable name `_id` confusing, because `_something` is usually a semi-private instance variable, not a local variable. Presumably this is to avoid shadowing python’s built-in `id` method? What about `rec_id` instead?

```suggestion
        rec_id = create_new_rec_id(phrase, speaker)
        audio_data.name = rec_id + ".wav"
        rec = Recording(
            id=rec_id,
            compressed_audio=audio_data,
            speaker=speaker,
            phrase=phrase,
            timestamp=0,
            comment=f"Uploaded by user: {request.user}",
            is_user_submitted=True,
        )
        rec.save()
        source = settings.RECVAL_AUDIO_PREFIX + rec_id + ".wav"
        dest = settings.RECVAL_AUDIO_PREFIX + rec_id + ".m4a"
```

validation/views.py
```diff
@@ -707,6 +734,59 @@ def save_wrong_word(request, recording_id):
...
+        subprocess.check_call(["ffmpeg", "-i", source, dest], cwd=settings.MEDIA_ROOT)
+        rec.compressed_audio = dest
+        rec.save()
+        os.remove(settings.MEDIA_ROOT + "/" + source)
```
andrewdotn: No harm in keeping the higher-quality original, just takes a bit of extra disk space.
```suggestion
```

validation/views.py
```diff
@@ -707,6 +734,59 @@ def save_wrong_word(request, recording_id):
...
+            speaker.save()
+
+        _id = create_new_rec_id(phrase, speaker)
+        audio_data.name = _id + ".wav"
+        rec = Recording(
+            id=_id,
+            compressed_audio=audio_data,
+            speaker=speaker,
+            phrase=phrase,
+            timestamp=0,
```
andrewdotn, line 771: If this is a float, could you store the actual current time, `time.time()` as a timestamp? The metadata of when the user-recording was uploaded will be handy in the future.

Actually I’d recommend:
  - Use a filename that is not a hash, e.g., `u${user.id}-${timestamp_with_millis}.wav`
  - Next to `foo.wav`, write out `foo.json` with all the original metadata for posterity: user-supplied transcript / translation, user id + login + full name, and timestamp

That way it’ll be much easier to potentially reuse these files for other things down the road. Basically anything you wished had just been in a JSON file next to every original audio file in the recording sessions, so you wouldn’t have had to parse it out of filenames or something: put all that in here from the start.

validation/views.py
```diff
@@ -835,3 +915,19 @@ def exclude_known_bad_recordings(recordings: QuerySet):
         # speaker shows up in API results.
         .exclude(speaker__gender__isnull=True)
     )
+
+
+def create_new_rec_id(phrase, speaker):
+    # Generate a unique ID for all user-submitted recordings
+    # Since these don't have a timestamp or a session,
+    # we use datetime.datetime.now().microsecond to add a truly unique element
```
```suggestion
    # we use time.time() to add a truly unique element
```

validation/views.py
```diff
@@ -46,6 +50,7 @@
 from django.contrib.auth.decorators import login_required
 from django.core.mail import mail_admins
 from django.db.models import Q, QuerySet
+from pydub import AudioSegment
```
andrewdotn: Is this used?

```suggestion
```

validation/models.py
```diff
@@ -83,6 +87,7 @@ class Phrase(models.Model):
         (LINKED, "Linked"),
         (VALIDATED, "Validated"),
         (REVIEW, "Needs review"),
+        (USER, "User-submitted"),
```
andrewdotn, line 90: I’m not sure what all of these statuses mean or when recordings should move from one status to another. Perhaps some documentation would clarify that?

andrewdotn (approved):
The only thing I would definitely want to see changed before merging this is to rename the migrations to not have `auto` in their names because that will be hard to fix later.

All other comments are optional at this time, as I know you want to get your demo working.

If you have any trouble with file permissions in production, let me know, as I’m happy to help.

validation/views.py
andrewdotn: I know we have these codes from importing data that originally was partly on paper. I know we need to keep these codes around to link to that. But I’d recommend a general change to speech-db where, if there’s a user object for any speaker, show “$username” in the interface instead of this speaker code.

If we’re gathering data with a computer into a relational database in the first place, we don’t need to keep using concepts invented as poor workarounds for not having a database, especially concepts that are both hard-to-understand and ambiguous.

validation/jinja2/validation/record_audio.html
nienna73: I do think having a pop-up is a good idea and I'll add it to the list for future developments

validation/jinja2/validation/record_audio.html
nienna73: I thought for some reason it was coming from npm, but this line clearly says that it isn't. Makes me wonder what [I installed](https://www.npmjs.com/package/recorderjs) then 🤔

validation/models.py
nienna73: I knew at one point, but now I have no clue again. I agree that it should be clearly outlined how things move from one status to another. I'll add this to my list and properly document it soon.

validation/static/record.js
nienna73: Sure does!

validation/views.py
nienna73: Nope, thanks!

validation/views.py
nienna73: This is something I was worried about before, yeah. The username is guaranteed to be unique, so I can use that as the speaker code for now at least.

validation/views.py
nienna73: Ooh okay, two things:
1. `timestamp` is a bit misleading: it's the time, in milliseconds, when this utterance occurs in the larger .wav file, not the time of day it was recorded
2. Yes. Yes please. This is my new priority with this project, second only to seeing if I can actually save these files on the server side.

validation/views.py
nienna73: Done, `rec_id` is much better.

validation/jinja2/validation/record_audio.html
andrewdotn: Oh then you’ll *probably* be able to reference the JS files from that somehow, but I’d have to poke a bit to see how to get that working here because right now nothing is doing that. It looks like tailwind gets installed by npm but has never actually been used?

nienna73: Yeah, I installed tailwind at some point, but didn't end up using it I guess. We aren't actually using Node, so that's probably why it isn't working.

PR #249, opened 2021-08-12
Add LOGIN_URL to settings

andrewdotn:
Right now, if you’re not logged in, and you click a link to a speech-db page that requires login, you get redirected to https://speech-db.altlab.app/accounts/login/?next=/whatever which is a 404 Not Found page.

This *should* set the correct thing in Django’s internals so that it redirects to the working `/login?next=/whatever` instead.

(I do not have python3.7 on the machine I am currently using, so cannot easily test.)


nienna73 (approved):
Beautiful! Thank you!

PR #252, opened 2021-08-12
Save Metadata

nienna73:
## What's in this PR:
* save the metadata for a user-submitted recording
* give a larger text area
* use "Cree" and "English" instead of "Transcription" and "Translation"
* pass `user.is_authenticated` to the new view so the options menu displays correctly

### DIFF
--- a/validation/views.py
+++ b/validation/views.py
@@ -964,4 +964,4 @@ def save_metadata_to_file(rec_id, user, transcription, translation):
         "dialect": "Plains Cree",
     }
     with open(dest, "w+") as f:
-        json.dump(data, f)
+        json.dump(data, f, indent=2, ensure_ascii=False)
--- a/validation/views.py
+++ b/validation/views.py
@@ -778,10 +778,22 @@ def record_audio(request):
         subprocess.check_call(["ffmpeg", "-i", source, dest], cwd=settings.MEDIA_ROOT)
         rec.compressed_audio = dest
         rec.save()
-        return HttpResponseRedirect("/secrets/record_audio")
+
+        context = dict(
+            form=form,
+            auth=request.user.is_authenticated,
+            is_linguist=user_is_linguist(request.user),
+        )
+        return HttpResponseRedirect("/secrets/record_audio", context)
     else:
         form = RecordNewPhrase()
-    return render(request, "validation/record_audio.html", {"form": form})
+
+    context = dict(
+        form=form,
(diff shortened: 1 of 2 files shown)


validation/views.py
```diff
@@ -928,3 +942,26 @@ def create_new_rec_id(phrase, speaker):
...
+        "user_id": user.id,
+        "username": user.username,
+        "full_name": user.first_name + " " + user.last_name,
+        "recorded_on": str(datetime.datetime.now()),
+        "transcription": transcription,
+        "translation": translation,
+        "dialect": "Plains Cree",
+    }
+    with open(dest, "w+") as f:
+        json.dump(data, f)
```
andrewdotn: Python’s default JSON output is all on one line and turns non-ascii characters into ugly unicode escapes like `itw\\u00eawina` so for anything a human might see I tend to do:

```suggestion
        json.dump(data, f, indent=2, ensure_ascii=False)
```

validation/views.py
```diff
@@ -928,3 +942,26 @@ def create_new_rec_id(phrase, speaker):
...
+        + "metadata/"
+        + rec_id
+        + ".json"
+    )
+    data = {
+        "audio_file_name": rec_id + ".wav",
+        "user_id": user.id,
+        "username": user.username,
+        "full_name": user.first_name + " " + user.last_name,
+        "recorded_on": str(datetime.datetime.now()),
```
andrewdotn: Sadly this leaves out the timezone information. I had to look up a way to include that:

```suggestion
        "recorded_on": str(datetime.datetime.now().astimezone()),
```

validation/views.py
```diff
@@ -928,3 +942,26 @@ def create_new_rec_id(phrase, speaker):
...
     )
     return sha256(signature.encode("UTF-8")).hexdigest()
+
+
+def save_metadata_to_file(rec_id, user, transcription, translation):
+    dest = (
+        settings.MEDIA_ROOT
+        + "/"
+        + settings.RECVAL_AUDIO_PREFIX
+        + "metadata/"
```
andrewdotn, line 952: Would this directory need to be created, or does it already exist?

andrewdotn (approved):
Two small concerns about the JSON format, but looking good!

validation/views.py
nienna73: I just made it on the server, so everything should be good to go now

PR #258, opened 2021-08-16
Small fixes to record audio

nienna73:
## What's in this PR:
* strip white space and newlines from transcription and translation
* remove the last 0.1 seconds from all recordings. This is the average time it takes someone to click down on their mouse, if I remember correctly
* also cleaned up the imports on another management command


aarppe:
I presume that we keep the original, but then remove the final 100ms from whatever is produced for actual use, e.g. in itwêwina?

andrewdotn (approved):
It might be nice to do the trimming client-side, so that the preview during a recording session plays back exactly how it’s going to sound once uploaded, but that’s probably way way more work.

nienna73:
> I presume that we keep the original, but then remove the final 100ms from whatever is produced for actual use, e.g. in itwêwina?

Yes, the original .wav file is kept, with the click at the end.

nienna73:
> It might be nice to do the trimming client-side

I agree, and you're correct. It would be great if the user could preview their recording as it will be saved into the DB, but until they hit save, that recording is just a byte stream/in memory file--it can't be properly opened and manipulated.

PR #260, opened 2021-08-30
Small bug fixes and improvements to linguist features

nienna73:
## What's in this PR:
Various bug fixes/improvements, as requested by @aarppe 
* fix: a speaker doesn't **have** to have a User
* remove the "Cancel" button when you're editing an entry (save or ignore form are the only options)
* preload the current phrase data to the Edit form
* show the analysis in the first table on the "More Options" page
* show who last edited an entry and when (also on the "More Options" page)
* when filtering by session, show entries from just that specific **session**, not all entries from that day
* fix: suggested analyses are of the inflected wordform, not of the lemma
* auto-load form data when resolving issues

## Notes:
For some Django forms, you can supply an `initial` argument that will populate the form data upon initialization. I can't figure out exactly which forms this does and does not work for, so for the forms used to resolve issues, I had to use JS to pre-populate those fields. It works, but there *should* be a better way to get it to work.

The new analysis function needs `crk-descriptive-analyzer.hfstol` to be present at the root of the project. The docker container is looking in the same place for `crk.zhfst`, so I just need to find how/where it's doing that and add the new file in the same way.


validation/jinja2/validation/view_issue_detail.html
```diff
@@ -19,6 +19,24 @@
...
+        </script>
+    {% elif issue.recording %}
+        <script>
+            document.addEventListener('DOMContentLoaded', () => {
+                let phrase = document.getElementById("id_phrase");
+                phrase.value = "{{ issue.suggested_cree }}"
+            })
+        </script>
+    {% endif %}
+
```
nienna73: This has to be done inline, otherwise it doesn't have access `issue` and fills in the literal string "{{ issue.etc }}". I don't love this solution, but it gets the job done. I'm more than willing to brainstorm a better way of accomplishing this.

validation/helpers.py
```diff
@@ -1,7 +1,8 @@
 from os import fspath
 
 import divvunspell
-from difflib import Differ, SequenceMatcher, get_close_matches
+from difflib import Differ
+import hfst_optimized_lookup
```
andrewdotn, line 5: This should be added to the Pipfile.

validation/jinja2/validation/segment_details.html
```diff
@@ -36,19 +36,22 @@
         <button class="button button--success button--back" onclick='goBack()'>Back</button>
     </div>
 
+    <p>This entry was last edited by {{ history[0].modifier }} on {{ history[0].history_date.strftime('%Y-%m-%d') }}</p>
```
andrewdotn: In case the `history` list is empty, this `<p>` should be wrapped in a conditional.

(I don’t know how often `history` might be empty, but in dev mode, `loaddata` seems to create entries without any history.)

validation/jinja2/validation/view_issue_detail.html
andrewdotn: [Form initial data](https://docs.djangoproject.com/en/3.2/ref/forms/api/#initial-form-values) will only be used if the form is not ‘bound’, i.e., if it appears to be a brand-new empty form created only by code, and not one already incorporating some user input.

[The current code in the `view_issue_detail` view](https://github.com/UAlbertaALTLab/recording-validation-interface/blob/b7948e304abbf4057fb221c5dca8cec62539f54c/validation/views.py#L548) does `form = EditIssueWithRecording(request.POST)` regardless of whether there is any POST data or not, meaning that django thinks this is never a brand-new form but always a user submission, which is why the `initial` value gets ignored.

I think this would be a more fruitful direction to go in:

```
def view_issue_detail(request, issue_id):
    issue = Issue.objects.get(id=issue_id)

    if issue.recording:
        if request.method == "POST":
            form = EditIssueWithRecording(request.POST)
            if form.is_valid():
                return handle_save_issue_with_recording(form, issue, request)
        else:
            form = EditIssueWithRecording(initial={
                'phrase': issue.comment
            })

    if issue.phrase:
        if request.method == 'POST':
            form = EditIssueWithPhrase(request.POST)
            if form.is_valid():
                return handle_save_issue_with_phrase(form, issue, request)
        else:
            form = EditIssueWithPhrase(initial={
                'transcription': issue.suggested_cree,
                'translation': issue.suggested_english
            })

    context = dict(
        issue=issue,
        form=form,
        auth=request.user.is_authenticated,
        is_linguist=user_is_linguist(request.user),
    )
    return render(request, "validation/view_issue_detail.html", context)
```

validation/forms.py
```diff
@@ -130,15 +130,17 @@ class Meta:
 
 class EditIssueWithPhrase(forms.ModelForm):
     transcription = forms.CharField(
-        max_length=200,
+        max_length=412,
```
andrewdotn, line 135: In other django projects, I’ve started changing these random length values to reference a `DEFAULT_MAX_LENGTH=500` constant at the top of the file.

andrewdotn (commented):
I hope my comments help?

aarppe:
> when filtering by session, show entries from just that specific session, not all entries from that day

To clarify, the intention is that we would show only words/sentences that were covered in an individual session, rather than smushing together the entire day, since the semantic topics in the concurrent or subsequent sessions can be divergent. However, I do consider as a useful feature showing in conjunction of words (and sentences) covered in some particular session recordings of these same words from other sessions.

validation/helpers.py
nienna73: `difflib` seems to come with python, but I've added `hfst-optimized-lookup`

validation/jinja2/validation/view_issue_detail.html
nienna73: Thank you for figuring this out! This is so much better than the JS.

andrewdotn (approved):

PR #284, opened 2021-10-29
fix: use new image retrieval method for individual speaker page

nienna73:
This method actually works in production, but I did sadly have to hard code a list of available images we have. The files live is drastically different places locally versus in production, so I tested this directly on production.


dwhieb (approved):
Well that's obnoxious. Looks good though!

PR #302, opened 2021-12-03
change: favicon

nienna73:
## What's in this PR:
* new icon for the tab (the "m" was for Maskwacîs)


dwhieb (approved):
Neat! I like the new one.

aarppe:
I thought it was a Dalek until I found the actual file. At the very least we do now not appear beholden to any single community, which is a good thing.

PR #318, opened 2021-12-18
feat: import synthesized recordings

nienna73:
Import the synthesized wâpamêw paradigm from Dolores


aarppe:
@nienna73 does this mean that I could (soon) listen to synthesized Dolores in speech-DB? (if not in itwêwina yet)

nienna73:
Yes, the synthesized wapamew paradigm should be up later today. None of them have transcriptions yet, but we can solve that problem later.

PR #390, opened 2022-04-29
Semantic class

nienna73:
## What's in this PR:
Add support for sorting by semantic class. Also add the elicitation sheets so we can find the semantic class for many of the entries.


aarppe:
Just so that we don't forget, we need to recall that for some MD entries (pre-existing in the MD content or recorded as "new") we might have two RW classifications - one resulting from the elicitation sheets and another from Daniel's classification of CW content. The latter might not cover entirely the former, and it is not guaranteed that they always agree, in terms of semantic domain and/or specificity.

nienna73:
Yes, I have it set up so that a phrase can have more than one semantic class and a semantic class can come from a number of places.

